PR #3, opened 2020-03-25
I forgot to commit along the way

GDemps:
Grant Dempsey exercise. Made it into Iteration 3, but didn't make it as far as would liked.

### DIFF
--- a/lib/ShoppingCart.rb
+++ b/lib/ShoppingCart.rb
@@ -1,0 +1,24 @@
+require './lib/product'
+
+class ShoppingCart
+
+attr_reader :name,
+            :capacity,
+            :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity
+    @products = []
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    total_number_of_products = @cart.products do |@cart.product|
+      @quantity.to_i
+  end
+
+end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,13 +1,35 @@
-class Prodct
+class Product
 
-  def initialize(category, name, unit_price, quantity)
+  attr_reader :category,
+              :name,
+              :unit_price,
+              :quantity
+
+  def initialize(category, name, unit_price = 3.70, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
   end
 
-  def total_price
-    @unit_price * @quantity
+  def quantity
+    @quantity.to_i
   end
+
+  def total_price
+    @quantity.to_f * @unit_price
+  end
+
+  def is_hoarded?
+    if @quantity == "0"
+      return true
+    else
+      return false
+    end
+  end
+
+  def hoard
+    @quantity = "0"
+  end
+
 end
--- a/test/product_test.rb
+++ b/test/product_test.rb
@@ -1,43 +1,46 @@
 require 'minitest/autorun'
 require 'minitest/pride'
 require './lib/product'
+require "pry"
 
 class ProductTest < Minitest::Test
+
   def test_it_exists
     product = Product.new(:paper, 'toilet paper', 3.70, '10')
     assert_instance_of Product, product
   end
 
   def test_it_has_a_category
-    skip
     product = Product.new(:paper, 'toilet paper', 3.70, '10')
     assert_equal :paper, product.category
   end
 
   def test_it_has_a_name
-    skip
     product = Product.new(:paper, 'toilet paper', 3.70, '10')
     assert_equal 'toilet paper', product.name
   end
 
   def test_it_has_a_unit_price
-    skip
     product = Product.new(:paper, 'toilet paper', 3.70, '10')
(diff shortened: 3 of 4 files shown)


lib/ShoppingCart.rb
```diff
@@ -0,0 +1,24 @@
+require './lib/product'
```
timomitchel: No need to require this file. We would only need it if we did a `Product.new` somewhere in this class.

lib/ShoppingCart.rb
```diff
@@ -0,0 +1,24 @@
...
+    @capacity = capacity
+    @products = []
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    total_number_of_products = @cart.products do |@cart.product|
```
timomitchel: This is on its way, and the right idea, but the syntax is a bit off. We don't have any instance variables in this class named `@cart`. Remember we cannot access instance variables from the tests by calling them in the class file, and vice versa. We can only access instance variables that we have defined above in initialize. The `sum` enumerable was the one that would best get the job done here, and we start on our `@products` array:
```
@products.sum do |product|
product.quantity
end
```

This is the same as doing something like this with each:
```
sum = 0
@products.each do |product|
sum += product.quantity
end
sum
```

lib/product.rb
```diff
@@ -1,13 +1,35 @@
-class Prodct
+class Product
 
-  def initialize(category, name, unit_price, quantity)
+  attr_reader :category,
+              :name,
+              :unit_price,
+              :quantity
+
+  def initialize(category, name, unit_price = 3.70, quantity)
```
timomitchel: We don't need unit_price set with a default variable, because we are always passing one in. So line 8 should read"
`def initialize(category, name, unit_price, quantity)`

lib/product.rb
```diff
@@ -1,13 +1,35 @@
...
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
   end
 
+  def quantity
```
timomitchel: We don't ever want to define a method that is the same name as an attr_reader. This will cause scope issues. You could do you string to integer conversion above right in initialize on line 12:
`@quantity = quantity.to_i`

lib/product.rb
```diff
@@ -1,13 +1,35 @@
...
+    @quantity.to_i
+  end
+
   def total_price
-    @unit_price * @quantity
+    @quantity.to_f * @unit_price
+  end
+
+  def is_hoarded?
+    if @quantity == "0"
```
timomitchel: This method was not based on a quantity, although I can see the logic there thinking about it in real world terms. This method should only have one job, return the state of an instance variable that you define and that begins as false. Something like: 
```
def is_hoarded?
@hoarded
end
```
Where `@hoarded` is initialized as false just underneath `@quantity` in the initialize method.
`@hoarded = false`
And then it is reassigned to true within the `hoard` method.

lib/product.rb
```diff
@@ -1,13 +1,35 @@
...
+      return true
+    else
+      return false
+    end
+  end
+
+  def hoard
+    @quantity = "0"
```
timomitchel: Based on the above comment, this method would read:
```
def hoard
@hoarded = true
end
```

test/product_test.rb
```diff
@@ -1,43 +1,46 @@
...
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
-
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_product_is_hoarded?
+    @product = Product.new(:paper, 'toilet paper', 3.70, '10')
+    binding.pry
```
timomitchel: Love to see you using pry. Keep pushing yourself to become fluent with this tool.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,36 @@
...
+require "pry"
+
+class ShoppingCartTest < Minitest::Test
+
+  def test_it_exists
+    cart = ShoppingCart.new("King Soopers", "30items")
+    assert_instance_of ShoppingCart, cart
+  end
+
+  def test_it_added_product
```
timomitchel: Great testing flow here.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  def test_if_total_number_of_products_works
+    cart = ShoppingCart.new("King Soopers", "30items")
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    binding.pry
+    assert_equal cart.total_number_of_products, 3
```
timomitchel: Great setup for this method, but the `assert_equal` should have read:
`assert_equal 13, cart.total_number_of_products`
(based on the interaction pattern)

GDemps:
Copy that.  I will run through it with a mentor.  Cheers!

PR #4, opened 2020-03-25
Rostam Mahabadi

Rostammahabadi:


lib/product.rb
```diff
@@ -1,13 +1,24 @@
-class Prodct
+class Product
```
memcmahon: The Product class looks great!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,67 @@
...
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    total_quantity = 0
+    @products.each do |product|
+      total_quantity += product.quantity
+    end
+    total_quantity
```
memcmahon: Great use of `each` here!  You could also use the enumerable `sum` 😉

lib/shopping_cart.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def is_full?
+    total_quantity = total_number_of_products
+    total_quantity >= @capacity
+  end
```
memcmahon: This definitely works, but you don't need to reassign the return value of the total_number method, you could use it like a variable:

```ruby
  def is_full?
    total_number_of_products >= @capacity
  end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def products_by_category(category)
+    categorical= []
+    @products.group_by do |product|
+      if product.category == category
+        categorical << product.category
+      end
+    end
+    categorical
+  end
```
memcmahon: You are using `group_by` but not appropriately.  The way you have this method written, you should use `each`.  Also, group_by would not help you for this method.  Also, we wanted this method to return and array of instances of products, not an array of categories.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def detail
+    details = Hash.new()
+    details[:name] = @name
+    details[:capacity] = @capacity
+    details
+  end
```
memcmahon: The logic totally works, and you could simplify it like this:

```ruby
def details
  { name: @name, capacity: @capacity }
end
```

But, the big issue is that we wanted this method to be called `details`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def percentage_occupied
+    occupied = total_number_of_products
+    return (occupied / @capacity.to_f).round(4)*100
```
memcmahon: similar to my note on `is_full` you don't _need_ to reassign the return value of the method to a local variable.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def sorted_products_by_quantity
+    quantity = @products.map.sort_by do |product|
+      product.to_s.length
```
memcmahon: This will sort products by the length of the string of their instance notation - we wanted it sorted by quantity, in descending order:

```ruby
@products.sort_by do |product|
  product.quantity
end.reverse

File: lib/shopping_cart.rb [hunk 8]
```diff
@@ -0,0 +1,67 @@
...
+
+  def product_breakdown
+    breakdown = Hash.new()
+    #issue: it is rewriting the :paper key when i need it to add it to
+    #the existing hash key :paper
+    @products.group_by do |product|
+      breakdown[product.category] = product
+    end
+
+  end
```
memcmahon: The reason you are having this issue is because you are not quite using `group_by` appropriately.  So, as you iterate, you are resetting the value at any give category-key to the product you are iterating over, instead of accumulating those products into an array based on their key.  Either of the below would work:

```ruby
  product.category
end
```

```ruby
breakdown = Hash.new
@products.each do |product|
  breakdown[product.category] = [] if breakdown[product.category].nil?
  breakdown[product.category] << product
end
breakdown
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  def test_cart_can_take_products
+    cart = ShoppingCart.new("King Soopers", "30items")
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+
+    cart.add_product(product1)
+    cart.add_product(product2)
+
+    assert_equal 2, cart.products.length
```
memcmahon: It would be better to be more explicit here - don't just verify two things are in the array, make sure they are the things that you expect:

```
assert_equal [product1, product2], cart.products
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,143 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product4)
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+
+    assert_equal 2, cart.products_by_category(:paper).length
```
memcmahon: Similar to my note above, be more specific!  This test is passing, but you are returning an array with the wrong elements in it.  The assertion below would help you fix the incorrect method output:

```ruby
assert_equal [product1, product3], cart.products_by_category(:paper)
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,143 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product4)
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+
+    assert_equal [:name, :capacity], cart.detail.keys
```
memcmahon: A better way to test this is to use the full hash as your expected value:

```
assert_equal ({name: 'King Soopers', capacity: 30}), cart.details
```

When you want to make an assertion on a hash, you need to wrap the hash in parens (or create an expected variable)

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,143 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product4)
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+
+    assert_equal :produce, cart.sorted_products_by_quantity[0].category
```
memcmahon: This passes, but does not confirm the functionality - when possible, assert based on the exact expected value.

PR #5, opened 2020-03-25
Ross Perry Shopping Cart Assessment

perryr16:


lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
+  attr_reader :category, :name, :unit_price, :quantity
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
+    @hoarded = false
```
aziobrow: these all look great, and nice job catching that the quantity needed to be an integer.

lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
     @unit_price * @quantity
   end
+
+  def is_hoarded?
+    @hoarded
+  end
+
+  def hoard
+    @hoarded = true
+  end
```
aziobrow: perfect for both of these

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
+
+class ShoppingCart
+
+
+  attr_reader :name, :capacity, :products
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+    @products = []
+  end
+
+  def add_product(product)
+    @products << product
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def total_number_of_products
+    sum = 0
+    @products.each do |product|
+      sum += product.quantity
+    end
+    sum
```
aziobrow: this is great! Ruby also has a perfect method for this that's designed to do exactly this job:

```
def total_number_of_products
    products.sum do |product|
        product.quantity
    end 
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def is_full?
+    total_number_of_products >= @capacity
```
aziobrow: nice! i like that you accomplished this in a single line. technically, the interaction pattern says that a cart is full if its total products _exceeds_ its capacity, which would only be greater than and not equal to.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def products_by_category(category)
+    products_by_category = @products.group_by do |product|
+      product.category
+    end
+    products_by_category[category]
```
aziobrow: cool! This definitely works, and I love that you're pushing yourself to use some tools outside of your comfort zone.

This use case is probably textbook for the use of `find_all` though (get a subset of the original collection that meet a certain criterion). So a more efficient alternative would be:

```
end 
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def details
+    details = {}
+    details[:name] = @name
+    details[:capacity] = @capacity
+    details
+    #hard coded?
```
aziobrow: a little bit, re: hardcoding, but in this case, it's appropriate. you're using dynamic values `(@name and @capacity)` to provide the details, which is the important part. (i.e., it would be bad to say `details[:name] = "King Soopers"`). 

the keys, however, are hardcoded, but some hashes will have static keys (as in this case, where you just have two that are given to you: name and capacity), in which case hardcoding is the only way to get the keys in there. Other hashes (like the `product_breakdown` of this class) are variable, and meant to be built dynamically (typically via iteration of some sort).

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def percentage_occupied
+    ((total_number_of_products.to_f / @capacity.to_f) * 100).round(2)
```
aziobrow: good!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def sorted_products_by_quantity
+    sorted_products_by_quantity = @products.sort_by do |product|
+      product.quantity * -1
```
aziobrow: nice! `sort_by` was the perfect tool here. 

this solution for changing the order is a little bit non-standard (although it definitely works).  You'd more commonly see something like a `.reverse` at the end, or the comparison (spaceship) operator (https://apidock.com/ruby/v2_5_5/Enumerable/sort).

aziobrow: You also don't need to save the result to a variable `sorted_products_by_quantity` since you're never actually using that variable anywhere.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+  end
+
+  def product_breakdown
+    product_breakdown = @products.group_by do |product|
+      product.category
+    end
+    product_breakdown.sort.to_h
```
aziobrow: cool! nice solution here.

test/product_test.rb
```diff
@@ -9,35 +9,41 @@ def test_it_exists
...
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_is_product_hoarded
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    assert_equal false, product1.is_hoarded?
+    product1.hoard
+    assert_equal true, product1.is_hoarded?
```
aziobrow: good- i like that this test shows progression to ensure dynamic functionality.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,144 @@
...
+    cart.add_product(product3)
+    cart.add_product(product4)
+
+    expected_collection = {:meat => [product2], :paper => [product1, product3], :produce => [product4]}
+
+    assert_equal expected_collection, cart.product_breakdown
+
+  end
+
+
```
aziobrow: tests look great. nice job following the interaction pattern exactly.

PR #6, opened 2020-03-25
Whitney Kidd

whitneykidd:


timomitchel:
Great job committing throughout.

lib/product.rb
```diff
@@ -1,13 +1,22 @@
-class Prodct
-
+class Product
+  attr_reader :category, :name, :unit_price, :quantity, :hoarded
```
timomitchel: No need for the `:hoarded` attr_reader here, because `is_hoarded?` essentially does that work for us.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+    @capacity = capacity.to_i
+    @products = []
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    @products.sum { |product| product.quantity }
```
timomitchel: The perfect enum for the job here.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def is_full?
+    @capacity < total_number_of_products ? true : false
```
timomitchel: Oh you fancy with your ternary ay? Still, it's overkill here. We only need the condition `@capacity < total_number_of_products`.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def products_by_category(category)
+    products_in_category = []
```
timomitchel: Great job using `each` to get the job done here. `find_all` would have saved you some work.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+    @products.each do |product|
+      if product.category == category
+        products_in_category << product
+      end
+    end
+    products_in_category
+  end
+
+  def details
+    cart_details = {:name => @name, :capacity => @capacity}
```
timomitchel: Nice job building the hash. This will run just fine without the local variable assignment.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def percentage_occupied
+    (( total_number_of_products.to_f / @capacity.to_f ) * 100).round(2)
```
timomitchel: Great job drawing on the return value of a method you previously built here!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def sorted_products_by_quantity
+    ( @products.sort_by { |product| product.quantity }).reverse
```
timomitchel: No need for the parentheses, but other than that exactly what we want.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def breakdown
+    product_breakdown = {}
+    @products.each do |product|
+      if product_breakdown.keys.include?(product.category)
```
timomitchel: Another way to write this condition that is a bit cleaner is: 
`if product_breakdown[product.category]`
Because it will be nil if not set yet. Also, although you built this correctly, you missed this one according to our tests, because you named the method wrong. The method should have been named `product_breakdown`!

test/product_test.rb
```diff
@@ -9,35 +9,38 @@ def test_it_exists
...
   def test_it_can_get_total_price
-    skip
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_can_be_hoarded
```
timomitchel: Nice setup and execution here.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,88 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/product'
+require './lib/shopping_cart'
+
+class ShoppingCartTest < Minitest::Test
+  def setup
```
timomitchel: Solid use of the `setup` method.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,88 @@
...
+    @cart.add_product(@product4)
+    assert_equal true, @cart.is_full?
+  end
+
+  def test_returns_product_by_category
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    @cart.add_product(@product4)
+    assert_equal [@product1, @product3], @cart.products_by_category(:paper)
```
timomitchel: You could make this more robust by testing each category. ;)

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,88 @@
...
+    assert_equal [@product4, @product1, @product2, @product3], @cart.sorted_products_by_quantity
+  end
+
+  def test_returns_product_breakdown
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    @cart.add_product(@product4)
+
+    breakdown = {
```
timomitchel: This is beautiful, it's a real shame you didn't name this method correctly.

PR #7, opened 2020-03-25
Michelle :) 

foleymichelle9:


lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
+    @is_hoarded = false
```
aziobrow: perfect! and great work seeing that you had to change the quantity to an integer here.

lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
     @unit_price * @quantity
   end
+
+  def is_hoarded?
+    @is_hoarded
+  end
+
+  def hoard
+    @is_hoarded = true
+  end
```
aziobrow: perfect for both of these methods

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
+class ShoppingCart
+
+  attr_reader :name, :capacity, :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+    @products = []
+  end
+
+  def add_product(product)
+    @products << product
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def total_number_of_products
+    product_total = @products.map do |product|
+      product.quantity
+    end
+      product_total.sum
```
aziobrow: this is great! that was a good use of map to get the quantities and then summing them at the end.  Remember-- `sum` is an enumerable on its own, so you could have used it instead of map for an even more efficient method. 

```
def total_number_of_products
    products.sum do |product|
        product.quantity
    end 
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def is_full?
+    return true if total_number_of_products >= 30
+    false
```
aziobrow: very cool! i like this early return statement here, and it works perfectly. 

Technically, you can also even do this in one line. 

```
def is_full?
    total_number_of_products >= @capacity.to_i
end 
```

Since that statement returns a boolean on its own, you don't technically need the if/else. if that statement is true, the whole method will return true, and vice versa.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def products_by_category(category)
+    @products.find_all do |product|
+      product.category == category
```
aziobrow: perfect use of find_all! Great work.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def details
+    details = Hash.new
+    details[:name] = name
+    details[:capacity] = capacity
+    details
```
aziobrow: perfect 🙂

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def percentage_occupied
+  ((total_number_of_products * 100) /  @capacity.to_f).round(2)
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def sorted_products_by_quantity
+    sorted = @products.sort_by { |product| product.quantity }
+      sorted.reverse
```
aziobrow: perfect! great job using `sort_by`-- that's definitely the best tool here, and the using reverse to switch up the order!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def product_breakdown
+    breakdown = Hash.new
+
```
aziobrow: It looks like you ran out of time, but you can do something like this here: 

```
def product_breakdown
     product_breakdown = {}
     @products.each do |product|
          product_breakdown[product.category] = products_by_category(product.category)
      end
    product_breakdown
end
```

test/product_test.rb
```diff
@@ -9,35 +9,43 @@ def test_it_exists
...
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_can_determine_if_product_is_hoarded
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+
+    assert_equal false, product.is_hoarded?
+    product.hoard
+    assert_equal true, product.is_hoarded?
```
aziobrow: good progression here to check for dynamic behavior

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,108 @@
...
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    assert_equal 13, cart.total_number_of_products
+    assert_equal false, cart.is_full?
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product4)
+    assert_equal [product1, product3], cart.products_by_category(:paper)
```
aziobrow: you could check another category here to make sure it works for non-paper products

PR #8, opened 2020-03-25
Alex Pariseau

arpariseau:


lib/product.rb
```diff
@@ -1,13 +1,24 @@
-class Prodct
+class Product
```
memcmahon: The Product class looks great!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,50 @@
+require './lib/product'
```
memcmahon: Since we don't create any products in this class, we don't need to require in this file.  I would only require files when you see that class being called directly in a file (`Product.new(...)` for example)

lib/shopping_cart.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    total_prod = 0
+    @products.map do |product| total_prod += product.quantity end
+    total_prod
```
memcmahon: Since you are leveraging an accumulator, `map` is not the best-choice - `each` would be better.  Or, you could use the enumerable `sum`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def is_full?
+    total_number_of_products >= @capacity
+  end
+
+  def products_by_category(category)
+    @products.find_all do |product| product.category == category end
```
memcmahon: Great use of `find_all`.  When using single line syntax, you should use `{` and `}` instead of `do` and `end`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def details
+    details = {name: @name, capacity: @capacity}
```
memcmahon: You don't _need_ the variable assignment here

lib/shopping_cart.rb
```diff
@@ -0,0 +1,50 @@
...
+  def sorted_products_by_quantity
+    @products.sort_by {|product| product.quantity}.reverse
+  end
+
+  def product_breakdown
+    categories = @products.map do |product| product.category end
+    breakdown = {}
+    categories.uniq.map do |category| breakdown[category] = products_by_category(category) end
+    breakdown
+  end
```
memcmahon: This totally works - and you could also use `group_by`:

```ruby
  product.category
end
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      @cart.add_product(product1)
+      @cart.add_product(product2)
+      @cart.add_product(product3)
+
+      assert_equal [product1, product3], @cart.products_by_category(:paper)
+    end
+
+    def test_details_access
+      cart = {name: "King Soopers", capacity: 30}
```
memcmahon: Great job figuring out assertions with hashes!

PR #9, opened 2020-03-25
Madeleine Halle IC2

madhalle:
Complete through iteration 3 and part of 4

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,4 +1,4 @@
-`
+
 class Product
   attr_accessor :category, :name, :unit_price, :quantity
   def initialize(category, name, unit_price, quantity)
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,13 +1,26 @@
-class Prodct
-
+`
+class Product
+  attr_accessor :category, :name, :unit_price, :quantity
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
   end
 
   def total_price
     @unit_price * @quantity
+  end
+
+  def is_hoarded?
+    if @quantity < 100
+      false
+    else
+      true
+    end
+  end
+
+  def hoard
+   @quantity += 100
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,0 +1,83 @@
+
+
+
+#
+# pry(main)> cart.sorted_products_by_quantity
+# #=> [#<Product:0x00007fccd2985f53...>, #<Product:0x00007fccd2985f51...>, #<Product:0x00007fccd2985f48...>, #<Product:0x00007fccd29b5720...>]
+#
+# pry(main)> cart.product_breakdown
+# #=> {:meat=>[#<Product:0x00007fccd2985f48...>], :paper=> [#<Product:0x00007fccd29b5720...>, #<Product:0x00007fccd2985f51...>], :produce=> [#<Product:0x00007fccd2985f53...>]}
+# ```
+
+
+class ShoppingCart
+  attr_accessor :name, :capacity, :products, :quantity
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
+    @products = []
+    @quantity = 0
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    @products.each do |product|
+      @quantity += product.quantity.to_i
+    end
+    @quantity
+  end
+
+  def is_full?
+    if total_number_of_products < @capacity
+      false
+    else
+      true
+    end
+  end
+
+  def products_by_category(specified_category)
+    one_category = []
+    @products.each do |product|
+      if product.category == specified_category
+        one_category << product
+      end
+    end
+    one_category
+  end
+
+  def percentage_occupied
+    ((total_number_of_products.to_f/@capacity.to_f)*100).round(2)
+  end
+
+  def sorted_products_by_quantity
+    by_quantity_low_to_high = []
+    sorted = []
+    @products.each do |product|
+      #require "pry"; binding.pry
+      product.quantity.to_i
+      #require "pry"; binding.pry
+      by_quantity_low_to_high << product.quantity
+    end
+    by_quantity_low_to_high.sort.each do |quantity|
+      @products.each do |product|
+        if product.quantity.to_i == quantity
+          sorted << product
+        end
+      end
+    end
+    sorted.uniq
+  end
+
+  def product_breakdown
+
+    @products.each_with_object({}) do |(category, type), category|
+      #require "pry"; binding.pry
+      (category[type[:meat]] ||= []) << product
+      (category[type[:paper]] ||= []) << product
+      (category[type[:produce]] ||= []) << product
+    end
+  end
+end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -61,7 +61,7 @@ class ShoppingCart
       #require "pry"; binding.pry
       by_quantity_low_to_high << product.quantity
     end
-    by_quantity_low_to_high.sort.each do |quantity|
+    by_quantity_low_to_high.sort.reverse.each do |quantity|
       @products.each do |product|
         if product.quantity.to_i == quantity
           sorted << product
@@ -71,13 +71,13 @@ class ShoppingCart
     sorted.uniq
   end
 
-  def product_breakdown
-
-    @products.each_with_object({}) do |(category, type), category|
-      #require "pry"; binding.pry
-      (category[type[:meat]] ||= []) << product
-      (category[type[:paper]] ||= []) << product
-      (category[type[:produce]] ||= []) << product
-    end
-  end
+  # def product_breakdown
+  #
+  #   @products.each_with_object({}) do |(category, type), category|
+  #     #require "pry"; binding.pry
+  #     (category[type[:meat]] ||= []) << product
+  #     (category[type[:paper]] ||= []) << product
+  #     (category[type[:produce]] ||= []) << product
+  #   end
+  # end
 end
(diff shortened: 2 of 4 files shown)


timomitchel:
Nice job committing throughout the challenge.

lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
+class Product
+  attr_accessor :category, :name, :unit_price, :quantity
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
```

lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
   end
 
   def total_price
     @unit_price * @quantity
   end
+
+  def is_hoarded?
```
timomitchel: This is not what we were looking for here, although I can see the logic for a real world example that would use `@quantity`. We wanted this method to return a value that could change in conjunction with the hoard method. The best way to accomplish this would be initialize and `@is_hoarded` instance variable below `@quantity = quantity.to_i` and set it to false by default there. Then the `hoard` method would set `@is_hoarded` to true, and this method would simply return the instance variable.
```
def is_hoarded?
@is_hoarded
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
+
+
+
+#
+# pry(main)> cart.sorted_products_by_quantity
```
timomitchel: Love seeing that you're putting the interaction pattern right here in the test. Try to remember to remove all comments before submitting.

lib/product.rb
```diff
@@ -1,13 +1,26 @@
-class Prodct
 
+class Product
+  attr_accessor :category, :name, :unit_price, :quantity
```
timomitchel: These should all be `attr_reader`. We want to expose as little as possible to the outside world.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+# #=> [#<Product:0x00007fccd2985f53...>, #<Product:0x00007fccd2985f51...>, #<Product:0x00007fccd2985f48...>, #<Product:0x00007fccd29b5720...>]
+#
+# pry(main)> cart.product_breakdown
+# #=> {:meat=>[#<Product:0x00007fccd2985f48...>], :paper=> [#<Product:0x00007fccd29b5720...>, #<Product:0x00007fccd2985f51...>], :produce=> [#<Product:0x00007fccd2985f53...>]}
+# ```
+
+
+class ShoppingCart
+  attr_accessor :name, :capacity, :products, :quantity
```
timomitchel: These should all be `attr_reader`. The interaction pattern only calls for us to do work from within this class, so we don't want to allow any changes to our attributes from outside of the class.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
+    @products = []
+    @quantity = 0
```
timomitchel: We don't need an instance variable here. We can set this variable locally in the only method that uses it `total_number_of_products`.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    @products.each do |product|
```
timomitchel: Nice job getting this to work with `each`. The `sum` enum can save you a few lines, and eliminate the need for a variable: 

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+      @quantity += product.quantity.to_i
+    end
+    @quantity
+  end
+
+  def is_full?
+    if total_number_of_products < @capacity
```
timomitchel: A couple of issues here. There is no need for an `if` statement. We simply need the conditional statement, which should read: 
Based on the exact wording of the interaction pattern, a shopping cart is full if it has more products than its capacity.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+      false
+    else
+      true
+    end
+  end
+
+  def products_by_category(specified_category)
+    one_category = []
+    @products.each do |product|
```
timomitchel: Nice job building this method with each. `find_all` would save you a few lines here: 
```
@products.find_all { |product| product.category == specified_category}
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+      if product.category == specified_category
+        one_category << product
+      end
+    end
+    one_category
+  end
+
+  def percentage_occupied
+    ((total_number_of_products.to_f/@capacity.to_f)*100).round(2)
```
timomitchel: Great job getting this to run, but always be thinking about readability here. How could you make this method more readable?

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def sorted_products_by_quantity
+    by_quantity_low_to_high = []
```
timomitchel: Nice job working through this. Our friend `sort_by` in conjunction with `reverse` could save you some time and effort here:
`@products.sort_by {|product| product.quantity }.reverse`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,83 @@
...
+      @products.each do |product|
+        if product.quantity.to_i == quantity
+          sorted << product
+        end
+      end
+    end
+    sorted.uniq
+  end
+
+  # def product_breakdown
```
timomitchel: I like where your thinking was heading here, but it's not as complex as all this. `group_by` will do a lot of the work for you: 
```
products.group_by do |product|
      product.category
    end
```

test/product_test.rb
```diff
@@ -9,35 +9,46 @@ def test_it_exists
...
-    skip
+
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_has_been_hoarded
```
timomitchel: Nice job testing both possible scenarios for this method that returns a boolean.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,82 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/shopping_cart'
+require './lib/product'
+
+
+class ShoppingCartTest < Minitest::Test
+  def setup
```
timomitchel: Nice use of the `setup` method here to save yourself some time.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    @cart = ShoppingCart.new("King Soopers", "30items")
+    @product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    @product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    @product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    @product4 = Product.new(:produce, 'apples', 0.99, '20')
+  end
+
+  def test_it_has_a_name
```
timomitchel: All of the attributes could be combined into a single test.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    assert_equal [@product1, @product2], @cart.products
+  end
+
+  def test_total_number_of_products
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    assert_equal 13, @cart.total_number_of_products
```
timomitchel: This test should end on line 39, based on the name of the test, then there should be a separate test for `is_full?`

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    @cart.add_product(@product4)
+    @cart.products
+    assert_equal true, @cart.is_full?
+  end
+
+  def test_products_by_category
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    assert_equal [@product1, @product3], @cart.products_by_category(:paper)
```
timomitchel: Nice job here, you should add an assert for each category to make this more robust.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def test_percentage_occupied
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    assert_equal 43.33, @cart.percentage_occupied
+  end
+
+  def test_sorted_by_quantity
+
```
timomitchel: Remember to clean up some of this white space.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    @cart.add_product(@product3)
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product4)
+    assert_equal [@product4, @product1, @product2, @product3] , @cart.sorted_products_by_quantity
+  end
+
+  def test_product_breakdown
+  skip
+    @cart.add_product(@product1)
```
timomitchel: We would only ever want to add each product once in this test setup.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    assert_equal 43.33, @cart.percentage_occupied
+  end
+
+  def test_sorted_by_quantity
+
+
+    @cart.add_product(@product1)
```
timomitchel: I see why you used the `.uniq` now. Because you added products (1 & 2) multiple times within your test setup. This setup caused you extra work, and could ultimately lead to an inaccurate method in some cases. Make sure you pay close attention to detail to the interaction pattern.

PR #11, opened 2020-03-25
Jane Greene

janegreene:
Independent challenge 2

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -3,7 +3,7 @@ class Product
               :name,
               :unit_price,
               :quantity,
-              :hoard #not sure about this
+              :hoard
 
   def initialize(category, name, unit_price, quantity)
     @category = category
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -2,16 +2,26 @@ class Product
   attr_reader :category,
               :name,
               :unit_price,
-              :quantity
+              :quantity,
+              :hoard #not sure about this
 
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
     @unit_price = unit_price
     @quantity = quantity.to_i
+    @hoard = false
   end
 
   def total_price
     @unit_price * @quantity.to_f
+  end
+
+  def is_hoarded?
+    @hoard
+  end
+
+  def hoard
+    @hoard = true
   end
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -7,11 +7,11 @@ class Product
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
-    @unit_price = 3.70
-    @quantity = quantity
+    @unit_price = unit_price
+    @quantity = quantity.to_i
   end
 
   def total_price
-    @unit_price * @quantity
+    @unit_price * @quantity.to_f
   end
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,8 +1,12 @@
-class Prodct
+class Product
+  attr_reader :category,
+              :name,
+              :unit_price,
+              :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
+    @name = name
     @unit_price = 3.70
     @quantity = quantity
   end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,0 +1,9 @@
+class ShoppingCart
+  attr_reader :name,
+              :capacity
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity
+  end
+end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -12,4 +12,10 @@ class ShoppingCart
   def add_product(product)
     @products << product
   end
+
+  def total_number_of_products
+    # @products.select do |product|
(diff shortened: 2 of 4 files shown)


lib/product.rb
```diff
@@ -1,13 +1,27 @@
-class Prodct
+class Product
+  attr_reader :category,
+              :name,
+              :unit_price,
+              :quantity,
+              :hoard
```
aziobrow: be careful with this-- this attr_reader is creating a method called `hoard` for the instance variable, but you also defined a method yourself called `hoard` which conflicts.  I thiiiink you end up okay on this one, but only on accident-- in general, you don't want to find yourself in a position where you have two methods named the same thing defined in a single class 🙂

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
+    @hoard = false
```
aziobrow: these are all great! nice catches on the mistakes as well as the conversion of the quantity.

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
   end
 
   def total_price
-    @unit_price * @quantity
+    @unit_price * @quantity.to_f
```
aziobrow: since the unit_price is already a float, you don't need to convert quantity (it doesn't hurt anything, but you don't need it).  only one of the operands in a mathematical statement need to be a float in order for the result to be a float.

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
+  end
+
+  def is_hoarded?
+    @hoard
```
aziobrow: great!

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
+  end
+
+  def hoard
+    @hoard = true
```
aziobrow: perfect

lib/shopping_cart.rb
```diff
@@ -0,0 +1,21 @@
+class ShoppingCart
+  attr_reader :name,
+              :capacity,
+              :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
```
aziobrow: this isn't a great idea here-- `capacity` is a string ("30items"), and even though it actually happens to work here, it's a little risky to try to convert a string to an integer when it has non-number characters-- you can get unexpected behavior.  a better bet would be to delete the word "items" out of the string and then convert it to an integer, which would look like this:  `@capacity = capacity.delete('items').to_i`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    # @products.select do |product|
+    #   @quanity.to_i += @quanity
+    # end
```
aziobrow: ahhh, i like where your head's at!  The idea of iterating over your products is great, but there are a couple execution pieces that are a little off here. 

Remember `select` is an alias for `find_all`, which returns to you a subset of the original collection that meets a certain criterion.  In this case though, you don't want only certain products, you want the quantities for each and every one of them.  Specifically, you want to add up each product's quantity.  So you have a couple options here-- here's one using `each`. 

```
def total_number_of_products
    total_quantity = 0
    products.each do |product|
        total_quantity += product.quantity
    end 
    total_quantity
end
```

Or the way I'd do it, with a nice Ruby enumerable that's designed to do exactly this job: 

```
def total_number_of_products
    products.sum do |product|
        product.quantity
    end 
end
```

test/product_test.rb
```diff
@@ -9,35 +9,39 @@ def test_it_exists
...
     assert_equal 9.0, product2.total_price
   end
+
+  def test_is_hoarded?
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+
+    assert_equal false, product.is_hoarded?
+
+    product.hoard
+    assert_equal true, product.is_hoarded?
```
aziobrow: great test!

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,40 @@
...
+
+class ProductTest < Minitest::Test
+
+  def test_it_exists
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    assert_instance_of ShoppingCart, cart
+    assert_equal 'King Soopers', cart.name
+    assert_equal 30, cart.capacity
+    assert_equal [], cart.products
```
aziobrow: love it

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def test_add_product
+    cart = ShoppingCart.new("King Soopers", "30items")
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+
+    cart.add_product(product1)
+    cart.add_product(product2)
+    assert_equal [product1, product2], cart.products
```
aziobrow: 👍

PR #12, opened 2020-03-25
Taylor Keller

takeller:
Turning in the shopping cart independent challenge

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,13 +1,23 @@
-class Prodct
+class Product
 
+  attr_reader :category, :name, :unit_price, :quantity
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
+    @is_hoarded = false
   end
 
   def total_price
     @unit_price * @quantity
   end
+
+  def hoard
+    @is_hoarded = true
+  end
+
+  def is_hoarded?
+    @is_hoarded
+  end 
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -19,5 +19,5 @@ class Product
 
   def is_hoarded?
     @is_hoarded
-  end 
+  end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,0 +1,8 @@
+class ShoppingCart
+
+  attr_reader :name, :capacity
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity[0..1].to_i
+  end
+end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -37,4 +37,8 @@ class ShoppingCart
   def percentage_occupied
     (@total_number_of_products.to_f/@capacity.to_f * 100).round(2)
   end
+
+  def sorted_products_by_quantity
+    sorted_products = @products.sort { |a,b| b.quantity <=> a.quantity}
+  end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -29,8 +29,12 @@ class ShoppingCart
   #   products_in_category
   # end
 
-  # Refactor using .find_all enumerable... Way easier 
+  # Refactor using .find_all enumerable... Way easier
   def products_by_category(category)
     products_in_category = @products.find_all { |product| product.category == category}
+  end
+
+  def percentage_occupied
+    (@total_number_of_products.to_f/@capacity.to_f * 100).round(2)
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -41,4 +41,15 @@ class ShoppingCart
   def sorted_products_by_quantity
     sorted_products = @products.sort { |a,b| b.quantity <=> a.quantity}
   end
+
+  def product_breakdown
+    products_hash = {}
+    product_categories = @products.map { |product| product.category }
+    product_categories.uniq!
+
+    product_categories.each do |category|
+      products_hash[category] = products_by_category(category)
+    end
+    products_hash
+  end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,19 +1,30 @@
 class ShoppingCart
 
-  attr_reader :name, :capacity, :products, :total_number_of_prodcuts
+  attr_reader :name, :capacity, :products, :total_number_of_products, :details
   def initialize(name, capacity)
     @name = name
     @capacity = capacity[0..1].to_i
     @products = []
-    @total_number_of_prodcuts = 0
+    @total_number_of_products = 0
+    @details = {name: @name, capacity: @capacity}
   end
 
   def add_product(product)
     @products << product
-    @total_number_of_prodcuts += product.quantity
+    @total_number_of_products += product.quantity
   end
 
   def is_full?
-    @total_number_of_prodcuts >= @capacity
+    @total_number_of_products >= @capacity
+  end
+
+  def products_by_category(category)
+    products_in_category = []
+    @products.each do |product|
+      if product.category == category
+        products_in_category << product
+      end
+    end
+    products_in_category
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,8 +1,13 @@
 class ShoppingCart
 
-  attr_reader :name, :capacity
+  attr_reader :name, :capacity, :products
   def initialize(name, capacity)
     @name = name
     @capacity = capacity[0..1].to_i
+    @products = []
+  end
+
+  def add_product(product)
+    @products << product 
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -18,13 +18,19 @@ class ShoppingCart
     @total_number_of_products >= @capacity
   end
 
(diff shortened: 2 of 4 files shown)


lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
+  attr_reader :category, :name, :unit_price, :quantity
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
+    @is_hoarded = false
```
aziobrow: these all look great, and nice job catching that the quantity needed to be an integer.

lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
     @unit_price * @quantity
   end
+
+  def hoard
+    @is_hoarded = true
+  end
+
+  def is_hoarded?
+    @is_hoarded
+  end
```
aziobrow: great job with both of these

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
+class ShoppingCart
+
+  attr_reader :name, :capacity, :products, :total_number_of_products, :details
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity[0..1].to_i
```
aziobrow: i love that you tried to separate the number from the word 'items' and this works fine for this use case.  However, since you're always grabbing the first two characters of the string, this implementation would get you into trouble if the cart had a capacity of anything other than double digits (less than 10 or more than 100), so this probably isn't the ideal solution.  I'd do something like this: 

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+    @products = []
+    @total_number_of_products = 0
```
aziobrow: this doesn't need to be an instance variable since all the work is being done inside a method that runs the calculation.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+    @details = {name: @name, capacity: @capacity}
```
aziobrow: same here, this doesn't need to be an instance variable. it can just be returned as a result of a `details` method.

aziobrow: In general, we prefer to keep things out of instance variables unless they really need to be there.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def add_product(product)
+    @products << product
+    @total_number_of_products += product.quantity
```
aziobrow: cool- this logic and implementation definitely makes sense, and works for its intended purpose.  It does deviate a bit from the intent posed by the interaction pattern-- if the `total_number_of_products` was intended to be updated by the `add_product` method, you would have seen it in the interaction pattern immediately under the first use of `add_product` as a return value.  

I definitely count your implementation as correct, but something to watch out for.

aziobrow: This is what I would have done: 

```
def total_number_of_products
    products.sum do |product|
        product.quantity
    end 
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def is_full?
+    @total_number_of_products >= @capacity
```
aziobrow: nice! i like that you accomplished this in a single line. technically, the interaction pattern says that a cart is full if its total products exceeds its capacity, which would only be greater than and not equal to. this would be a great edge case to both catch and test.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+  #     if product.category == category
+  #       products_in_category << product
+  #     end
+  #   end
+  #   products_in_category
+  # end
+
+  # Refactor using .find_all enumerable... Way easier
+  def products_by_category(category)
+    products_in_category = @products.find_all { |product| product.category == category}
```
aziobrow: nice! love that you did it with each and then realized that there was a better way. perfect.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def percentage_occupied
+    (@total_number_of_products.to_f/@capacity.to_f * 100).round(2)
```

Small point, but you only need to convert either the numerator or denominator to a float in order for the output to be a float-- both is unnecessary.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def sorted_products_by_quantity
+    sorted_products = @products.sort { |a,b| b.quantity <=> a.quantity}
```
aziobrow: great! you could also have used sort_by and then reversed. 

`@products.sort_by(&:quantity).reverse`

aziobrow: Also, you don't need to save this result to a variable here, since you're never actually using it.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def product_breakdown
+    products_hash = {}
```
aziobrow: try not to name things with their data type...

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+    product_categories = @products.map { |product| product.category }
+    product_categories.uniq!
```
aziobrow: you don't need these two lines (47-48)

lib/shopping_cart.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    product_categories.each do |category|
+      products_hash[category] = products_by_category(category)
+    end
+    products_hash
+  end
```
aziobrow: this is exactly the right idea-- love that you're using your helper method `products_by_category`.  

I'd do this whole method with a slight variant on these few lines: 

```
def product_breakdown
     breakdown = {}
     @products.each do |product|
          breakdown[product.category] = products_by_category(product.category)
      end
     breakdown
end
```

There's also a great little enumerable called `reduce` that you can look into if you'd like, but we'll get into that as a class a little later in the mod.

test/product_test.rb
```diff
@@ -3,41 +3,45 @@
...
+    assert_equal false, @product1.is_hoarded?
+    assert_equal false, @product2.is_hoarded?
+
+    @product1.hoard
+    @product2.hoard
 
-    assert_equal 37.0, product1.total_price
-    assert_equal 9.0, product2.total_price
+    assert_equal true, @product1.is_hoarded?
+    assert_equal true, @product2.is_hoarded?
```
aziobrow: i'd definitely rather you over-test than under-test, but you can safely assume that, given your implementation, if it works for one product, it will work for any other.. if there was any sort of conditional in the method that handled different types of products differently, then I would expect to see different products tested, but that's not the case here.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  def test_add_product
+    assert_equal [], @cart.products
+
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+
+    assert_equal 2, @cart.products.length
+    assert_equal @product1, @cart.products[0]
+    assert_equal @product2, @cart.products[1]
```
aziobrow: this is fine, but i think this is more succinctly and readably accomplished by just asserting the array itself. 

`assert_equal [@product1, @product2], @cart.products`

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,111 @@
...
+    assert_equal 0, @cart.total_number_of_products
+
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+
+    assert_equal 12, @cart.total_number_of_products
+  end
+
+  def test_is_full?
+    assert_equal [], @cart.products
```
aziobrow: you don't need this assertion in this test-- you already test it as part of the `add_products` method test.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,111 @@
...
+    assert_equal false, @cart.is_full?
+
+    @cart.add_product(@product4)
+
+    assert_equal true, @cart.is_full?
+  end
+
+  def test_cart_details
+    cart_details = {name: "King Soopers", capacity: 30}
+    assert_equal cart_details , @cart.details
```
aziobrow: 👍

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def test_products_by_category
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    @cart.add_product(@product4)
+
+    assert_equal [@product1, @product3], @cart.products_by_category(:paper)
+    assert_equal [@product4], @cart.products_by_category(:produce)
```
aziobrow: love that you tested multiple categories here to ensure dynamic functionality.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,111 @@
...
+    product_hash = {meat: [@product2], paper: [@product1, @product3], produce: [@product4]}
+
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    @cart.add_product(@product4)
+
+    assert_equal product_hash, @cart.product_breakdown
+  end
+end
```
aziobrow: tests look great!

takeller:
Thanks so much for all of your feedback, Aurora! Having you look over my solutions is very helpful. I went back over everything to update my classes with your suggestions, and I'll try and keep in mind all of the points you made going forward.

PR #13, opened 2020-03-25
Jack Cullen

jpc20:


lib/product.rb
```diff
@@ -1,13 +1,26 @@
-class Prodct
-
-  def initialize(category, name, unit_price, quantity)
+class Product
+  attr_reader :category,
+              :name,
+              :unit_price,
+              :quantity,
+              :hoarded
```
memcmahon: Since we have an explicit method created below to see the value of our @hoarded variable, we do not need an attr_reader for it.

lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
+  def initialize(category, name, unit_price=3.70, quantity)
```
memcmahon: We don't need a default value for unit_price, in this case :)

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
+class ShoppingCart
+  attr_reader :name,
+              :capacity,
+              :products,
+              :details
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
+    @products = []
+    @details = {}
```
memcmahon: If you are going to make details an instance variable, you should just create it here, rather than relying on another method to populate the information.  But, we generally want to avoid unnecessary instance variables, so creating a `details` method would be best.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    @products.reduce(0) do |total, product|
```
memcmahon: Great use of `reduce` - you could also use the enumerable `sum`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+  def products_by_category(category)
+    @products.find_all do |product|
+      product.category == category
+    end
+  end
+
+  def generate_details
+    @details[:name] = @name
+    @details[:capacity] = @capacity
+  end
```
memcmahon: Based on the interaction pattern, we should not need to call `generate_details` or any other method, before calling `details`.  So, strictly speaking, the `details` method isn't working 😢 Follow the interaction pattern very closely!!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def percentage_occupied
+    ((total_number_of_products.to_f / @capacity) * 100).round(2)
+  end
+
+  def sorted_products_by_quantity
+    @products.sort_by do |product|
+      -product.quantity
```
memcmahon: yay, math

lib/shopping_cart.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+  end
+
+  def product_breakdown
+    @products.group_by do |product|
```
memcmahon: Great use of `group_by`!

test/product_test.rb
```diff
@@ -9,35 +9,47 @@ def test_it_exists
...
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_is_horded_false_by_default
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+
+    assert_equal :paper, product.category
+    assert_equal 37.0, product.total_price
```
memcmahon: Since we have confirmed these two methods above, we don't need to reassert them

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+
+    assert_equal false, @cart.is_full?
+
+    @cart.add_product(@product4)
+
+    assert_equal true, @cart.is_full?
+  end
```
memcmahon: You could put all of this in one test - assert on the falseness, add more products, and then assert on the trueness.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    assert_equal [@product1, @product3], @cart.products_by_category(:paper)
+  end
+
+  def test_cart_details
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+    @cart.add_product(@product4)
+
+    @cart.generate_details
```
memcmahon: This never appears in the interaction pattern!! Your test is giving you a false positive.

PR #14, opened 2020-03-25
Kiera Allen

KieraAllen:


lib/product.rb
```diff
@@ -1,13 +1,27 @@
-class Prodct
+class Product
+  attr_reader :category,
+              :name,
+              :unit_price,
+              :quantity
```
aziobrow: 👍

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
```
aziobrow: nice catches

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
     @quantity = quantity
```
aziobrow: according to the interaction pattern, quantity needed to be an integer, but it's being passed in as a string. So I would have expected to see some conversion here 

`@quantity = quantity.to_i`

aziobrow: Also, i have no idea why this comment is taking up the full width of the screen 🤷‍♀

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
+    @is_hoarded = false
   end
 
   def total_price
-    @unit_price * @quantity
+    @unit_price * (@quantity).to_f
```
aziobrow: nice conversion here, but as I mentioned above, the test says that `quantity` itself should be an integer, so if you'd done the conversion above, you wouldn't have needed it here.

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
   end
+
+  def is_hoarded?
+    @is_hoarded
```
aziobrow: beautiful

lib/product.rb
```diff
@@ -1,13 +1,27 @@
...
+  end
+
+  def hoard
+    @is_hoarded = true
```
aziobrow: perfect

lib/shopping_cart.rb
```diff
@@ -0,0 +1,12 @@
+class ShoppingCart
+  attr_reader :name
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity
```
aziobrow: you would have wanted `@products = []` here too, according to the interaction pattern.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  def capacity
+    @capacity.to_i
```
aziobrow: almost! but since capacity starts off as `"30items"` in our test example, you'd have to figure out how to get rid of the `items` in that string before you can convert to an integer.  Something like this: 

`@capacity.delete('items').to_i`

aziobrow: I'd also probably just take care of this when this instance variable is set in initialize as a matter of personal preference, instead of writing this method, but it works just fine here!

test/product_test.rb
```diff
@@ -3,41 +3,49 @@
...
+
+    assert_equal 3.70, @product1.unit_price
   end
 
   def test_it_has_a_quantity
-    skip
-    product = Product.new(:paper, 'toilet paper', 3.70, '10')
-    assert_equal 10, product.quantity
+
+    assert_equal "10", @product1.quantity
```
aziobrow: You changed the test here so that it's "10" instead of 10, which breaks from the interaction pattern.  The tests should exactly match the interaction pattern and should be the source of truth-- they should not be changed unless you've really thought through it and realized that you made an actual mistake.

test/product_test.rb
```diff
@@ -3,41 +3,49 @@
...
 
   def test_it_can_get_total_price
-    skip
-    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
-    product2 = Product.new(:meat, 'chicken', 4.50, '2')
 
-    assert_equal 37.0, product1.total_price
-    assert_equal 9.0, product2.total_price
+    assert_equal 37.0, @product1.total_price
+    assert_equal 9.0, @product2.total_price
```
aziobrow: 👍

test/product_test.rb
```diff
@@ -3,41 +3,49 @@
...
+
+  def test_that_is_hoarded_default_is_false
+
+    assert_equal false, @product1.is_hoarded?
+  end
+
+  def test_it_is_hoarded
+    @product1.hoard
+
+    assert_equal true, @product1.is_hoarded?
```
aziobrow: i'd probably put these both in the same test, since they're testing the same process.  Also, it proves that you built dynamic functionality since you're testing the exact same product over time. 

```
def test_it_can_be_hoarded 
    assert_equal false, @product1.is_hoarded? 

    @product1.hoard 

     assert_equal true, @product1.is_hoarded?
end
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,29 @@
...
+class ShoppingCartTest < Minitest::Test
+  def setup
+    @product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    @product2 = Product.new(:meat, 'chicken', 4.50, '2')
+
+    @cart = ShoppingCart.new("King Soopers", "30items")
+  end
+
+  def test_it_exists
+
```
aziobrow: tiny point, but you don't need these empty lines here.

KieraAllen:
Hey Aurora, thanks for all your edits! Your comments and working through this problem in a small group with Megan definitely helped my understanding, but I am interested in pairing with Graham. I feel like I should know who he is, but I don't, so can you send me his Slack handle?

PR #15, opened 2020-03-25
Jack Gould

jhgould:


timomitchel:
Nice job committing throughout!

lib/product.rb
```diff
@@ -1,13 +1,37 @@
...
+  attr_reader :category, :name, :unit_price, :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_f
```
timomitchel: Attention to detail here. The interaction pattern wanted this value to be an integer.

lib/product.rb
```diff
@@ -1,13 +1,37 @@
...
   end
 
   def total_price
     @unit_price * @quantity
   end
+
+  def is_hoarded?
```
timomitchel: So this isn't quite what we wanted, but I can see how your logic follows a real world scenario. We wanted this method to return a value that could change in conjunction with the hoard method. The best way to accomplish this would be initialize and @is_hoarded instance variable below @quantity = quantity and set it to false by default there. Then the hoard method would set @is_hoarded to true, and this method would simply return the instance variable.
```
def is_hoarded?
@is_hoarded
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,40 @@
+#shopping cart class
+require './lib/product'
```
timomitchel: No need to require this file, because we never explicitly refer to the Product class in this class. We pass in, hold onto, and manipulate Product objects, but we instantiate them in the test, so ShoppingCart doesn't need to know about this file, because when it gets passed objects, they are already fully formed. Only if we called Product.new(..) in this file, would we need to require the product file.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,40 @@
...
+  attr_reader :name, :capacity, :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
+    @products = []
+  end
+
+  def add_product(product)
+    if product.is_hoarded? == false
```
timomitchel: You're doing a lot of extra work here that the interaction pattern does not call for. This method should read: 
```
def add_product(product)
@products << product
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,40 @@
...
+        @products << product
+    else
+      p "this product has been hoarded"
+    end
+  end
+
+  def total_number_of_products
+    total = 0
+    @products.each do |i|
```
timomitchel: Nice job using each to get the job done. Ruby convention asks that we name our block variable in a way that it represents each element in the array. `i` is used in other languages, but in ruby we are more descriptive. The block variable should be `|product|` to represent a single `product` instance. And, there is an enum built specifically for this job named `sum`: 
```
@products.sum do |product|
product.quantity
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,40 @@
...
+    total += i.quantity
+    end
+    total
+  end
+
+  def is_full?
+    if total_number_of_products >= 30
```
timomitchel: No need for this if statement. We only need the condition. But, attention to detail here, because the interaction pattern said that a shopping cart is full if it has more products in it than its capacity. So we wanted:

test/product_test.rb
```diff
@@ -9,35 +9,50 @@ def test_it_exists
...
-    skip
+
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_is_not_hoarded
```
timomitchel: Nice job testing both a true and false scenario for this boolean method

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,50 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    cart = ShoppingCart.new("King Soopers", "30items")
+    cart.add_product(product1)
+    cart.add_product(product2)
+
+    assert_equal 20, cart.total_number_of_products
+  end
+
+  def test_it_is_full
```
timomitchel: Whenever we have a method that returns a boolean we want to make sure we test both the scenario that makes the method return a true and the scenario that makes it return a false. This only tests one scenario.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,50 @@
...
+    assert_equal "King Soopers", cart.name
+    assert_equal 30, cart.capacity
+    assert_equal [], cart.products
+  end
+
+  def test_it_can_add_product
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    assert_equal [product1], cart.add_product(product1)
```
timomitchel: Next time make sure to test methods like this exactly as the interaction pattern calls for, so in this case with two products passed in.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def test_it_has_total_number_of_products
+
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    cart = ShoppingCart.new("King Soopers", "30items")
+    cart.add_product(product1)
+    cart.add_product(product2)
+
+    assert_equal 20, cart.total_number_of_products
```
timomitchel: Nice job testing here, but once again, make sure to follow the interaction pattern exactly, which called for one more product in our array and resulted in a 13 for the total_number_of_products

PR #16, opened 2020-03-25
Jack Puchalla

JPuchalla:


lib/product.rb
```diff
@@ -1,13 +1,28 @@
...
+class Product
+  attr_reader :category, :name, :unit_price, :quantity, :total_price
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
+    @hoard = false
```
aziobrow: nice job with all of these!

lib/product.rb
```diff
@@ -1,13 +1,28 @@
...
+  end
+
+
+  def quantity
+    @quantity.to_i
```
aziobrow: you should be careful here--- I thiiiink you got away with this, but you have this method _and_ and attr_reader for quantity, which means you technically created two methods with the same name in the same class, which you would never want.  This one is the correct one.

lib/product.rb
aziobrow: I'd probably have done `@quantity = quantity.to_i` here.

lib/product.rb
```diff
@@ -1,13 +1,28 @@
...
   end
 
   def total_price
-    @unit_price * @quantity
+    @unit_price * @quantity.to_f
```
aziobrow: if you had changed quantity to an integer in initialize, you wouldn't have had to change this method at all.

lib/product.rb
```diff
@@ -1,13 +1,28 @@
...
   end
+
+  def is_hoarded?
+    @hoard
+  end
+
+  def hoard
+    @hoard = true
+  end
```
aziobrow: nice job with these!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,47 @@
+class ShoppingCart
+  attr_reader :name, :capacity, :products
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity
```
aziobrow: not quite. you'll notice that in the interaction pattern, capacity is being passed in as "30items" but it's supposed to be returned as the integer 30. So this doesn't quite do what it needs to do.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+
+  def add_product(item)
+    @products << item
+  end
+
+  def total_number_of_products
+     total_products = 0
+     @products.each {|product| total_products += product.quantity}
```
aziobrow: great! you could also have used `sum` here instead of `each` and then you could have deleted lines 19 and 21 and done this all in one line.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,47 @@
...
+     total_products
+  end
+
+  def is_full?
+    total = total_number_of_products
+    if total >= 30
```
aziobrow: this isn't quite right.  not all carts have the same capacity, and you're always checking for 30 here.  So if a cart had a capacity of, let's say, 10 items, or 50 items, this method wouldn't work.   You should use `@capacity` instead of hardcoding the number 30 here.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  def products_by_category(category)
+    product_with_category = []
+    @products.flat_map {|product| if product.category == category
+      product_with_category << product
+      end
+    }
+    product_with_category
```
aziobrow: this works! good job finding a working solution.  If you find yourself using an `if` statement inside of a map, looking for a certain condition, you probably mean `find_all`.   This is a great use case for `find_all`. 

```
end 
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def details
+   cart_details = {name: @name, capacity: @capacity}
+   cart_details
```
aziobrow: this is almost right, except your capacity wasn't quite right (remember, you have yours saved as '30items' the string, not 30 the integer.  But the way you built this hash is correct.

test/product_test.rb
```diff
@@ -3,41 +3,46 @@
...
-    assert_equal 9.0, product2.total_price
+    assert_equal 37.0, @product.total_price
+    assert_equal 9.0, @product2.total_price
+  end
+
+  def test_if_product_is_hoarded
+    
+    assert_equal false, @product.is_hoarded?
+    @product.hoard
+    assert_equal true, @product.is_hoarded?
```
aziobrow: very nice test for this

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,69 @@
...
+  def test_it_can_get_attributes
+    assert_equal 'King Soopers', @cart.name
+    assert_equal 30, @cart.capacity
+    assert_equal [], @cart.products
+  end
+
+
+  def test_it_can_add_product_to_cart
+    assert_equal [@product1], @cart.add_product(@product1)
+    assert_equal [@product1, @product2], @cart.add_product(@product2)
```
aziobrow: it's fine that you did assertions for these two lines (28 and 29), but you didn't have to.  If you look at the interaction pattern, we don't really care what the return value for the `add_product` is... we just care that in the end, when we check the `cart.products`, that we see the products we added.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,69 @@
...
+    assert_equal [@product1, @product2], @cart.products
+  end
+
+  def test_if_cart_can_measure_total_number_of_products
+    assert_equal [@product1], @cart.add_product(@product1)
+    assert_equal [@product1, @product2], @cart.add_product(@product2)
+    assert_equal [@product1, @product2, @product3], @cart.add_product(@product3)
```
aziobrow: same thing-- we don't really need assertions for these 3 lines (34, 35, 36)

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,69 @@
...
+    assert_equal 13, @cart.total_number_of_products
+  end
+
+  def test_if_cart_is_full
+
+    assert_equal [@product1], @cart.add_product(@product1)
+    assert_equal [@product1, @product2], @cart.add_product(@product2)
+    assert_equal [@product1, @product2, @product3], @cart.add_product(@product3)
+    assert_equal false, @cart.is_full?
+    assert_equal [@product1, @product2, @product3, @product4], @cart.add_product(@product4)
```
aziobrow: you're missing an assertion here-- you asserted that it wasn't full here on line 45, and then added a product-- i would have expected to see a new assertion now saying that it is full. Something like this: 

```
def test_if_cart_is_full 
    cart.add_product(@product1)
    cart.add_product(@product2)
    cart.add_product(@product3)

    assert_equal false, cart.is_full?

    cart.add_product(@product4)

    assert_equal true, cart.is_full?
end 
```

This is also what I mean about not testing the return value of adding products.  See how I'm not asserting the return value for every time i call `add_product`?

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  end
+
+  def test_if_it_can_retrieve_cart_details
+    assert_equal [@product1], @cart.add_product(@product1)
+    assert_equal [@product1, @product2], @cart.add_product(@product2)
+    assert_equal [@product1, @product2, @product3], @cart.add_product(@product3)
+    assert_equal [@product1, @product2, @product3, @product4], @cart.add_product(@product4)
+
+    assert_equal ({:name=>"King Soopers", :capacity=>"30items"}), @cart.details
```
aziobrow: this doesn't quite match the interaction pattern. make your tests match the interaction pattern... 

```
pry(main)> cart.details
#=> {:name=> "King Soopers", :capacity=> 30}
```

PR #17, opened 2020-03-25
Stella Bonnie 

stellakunzang:
Completed through Iteration 4 on Independent Challenge

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,13 +1,14 @@
-class Prodct
+class Product
+  attr_reader :category, :name, :unit_price, :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
   end
 
   def total_price
-    @unit_price * @quantity
+    @unit_price * (@quantity.to_i)
   end
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -6,9 +6,19 @@ class Product
     @name = name
     @unit_price = unit_price
     @quantity = quantity
+    @hoard = false
   end
 
   def total_price
     @unit_price * (@quantity.to_i)
+  end
+
+  def hoard
+    @quantity = @quantity.to_i * 3
+    @hoard = true
+  end
+
+  def is_hoarded?
+    @hoard
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -50,6 +50,10 @@ class ShoppingCart
   end
 
   def product_breakdown
-
+    product_breakdown = {}
+      @products.each do |product|
+        product_breakdown[product.category] = products_by_category(product.category)
+      end
+    product_breakdown
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -43,4 +43,13 @@ class ShoppingCart
   def percentage_occupied
     ((total_number_of_products / capacity.to_f) * 100).round(2)
   end
+
+  def sorted_products_by_quantity
+    sorted_products = @products.sort_by { |product| product.quantity.to_i }
+    sorted_products
+  end
+
+  def product_breakdown
+
+  end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -37,7 +37,6 @@ class ShoppingCart
     details[:name] = name
     details[:capacity] = capacity.to_i
     return details
-     #returns a hash with name and capacity as keys and strings as values
   end
 
   def percentage_occupied
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -45,7 +45,7 @@ class ShoppingCart
 
   def sorted_products_by_quantity
     sorted_products = @products.sort_by { |product| product.quantity.to_i }
-    sorted_products
+    sorted_products.reverse
   end
 
   def product_breakdown
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,0 +1,9 @@
+class ShoppingCart
+  attr_reader :name, :capacity, :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.delete('items')
+    @products = []
+  end
+end 
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -27,11 +27,16 @@ class ShoppingCart
     end
   end
 
-  def products_by_category
-
+  def products_by_category(category)
+    products_by_category = products.group_by { |product| product.category }
+    products_by_category[category]
   end
 
   def details
-
+    details = {}
+    details[:name] = name
+    details[:capacity] = capacity.to_i
+    return details
+     #returns a hash with name and capacity as keys and strings as values
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,9 +1,37 @@
 class ShoppingCart
-  attr_reader :name, :capacity, :products
+  attr_reader :name, :capacity, :products, :total_number_of_products
 
   def initialize(name, capacity)
     @name = name
     @capacity = capacity.delete('items')
     @products = []
+    @total_number_of_products = 0
   end
-end 
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    @products.each do |product|
+      @total_number_of_products += product.quantity.to_i
+    end
+    @total_number_of_products
+  end
+
+  def is_full?
+    if total_number_of_products >= @capacity.to_i
+      true
+    else
+      false
+    end
+  end
+
+  def products_by_category
+
+  end
+
+  def details
+
+  end
+end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -39,4 +39,8 @@ class ShoppingCart
     return details
      #returns a hash with name and capacity as keys and strings as values
   end
+
+  def percentage_occupied
+    ((total_number_of_products / capacity.to_f) * 100).round(2)
+  end
 end
--- a/test/product_test.rb
+++ b/test/product_test.rb
@@ -36,4 +36,15 @@ class ProductTest < Minitest::Test
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_can_hoard
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+    assert_equal false, product.is_hoarded?
+
+    product.hoard
+    assert_equal 30, product.quantity
+    assert_equal true, product.is_hoarded?
+  end
+
+
 end
--- a/test/product_test.rb
+++ b/test/product_test.rb
@@ -9,31 +9,27 @@ class ProductTest < Minitest::Test
   end
 
   def test_it_has_a_category
-    skip
     product = Product.new(:paper, 'toilet paper', 3.70, '10')
     assert_equal :paper, product.category
   end
 
   def test_it_has_a_name
-    skip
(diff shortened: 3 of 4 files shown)


stellakunzang:
I was being super consistent with tense in commits and then I got excited toward the end and started using past tense...dang!

lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
+  attr_reader :category, :name, :unit_price, :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
+    @hoard = false
```
aziobrow: great job on most of this! The only hitch is that you missed quantity here-- the interaction pattern and the test are asking for quantity to be an integer, and since it's being passed in as a string, I would expect to see some conversion here-- something like `@quantity = quantity.to_i`

lib/product.rb
```diff
@@ -1,13 +1,24 @@
-class Prodct
+class Product
+  attr_reader :category, :name, :unit_price, :quantity
```
aziobrow: nice job not including `hoard` as an attr_reader here, since you have the `is_hoarded?` method.

lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
+    @hoard = false
   end
 
   def total_price
-    @unit_price * @quantity
+    @unit_price * (@quantity.to_i)
```
aziobrow: if you'd done the conversion in initialize, this method would just work without needing any changes

lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
+  end
+
+  def hoard
+    @quantity = @quantity.to_i * 3
```
aziobrow: not sure where this line is coming from-- was it your idea of what this method should do? If so, i love it-- but for the purposes of an independent challenge, stick to the interaction pattern for implementation 🙂 So in that case, you'd only need the next line, line 18.

lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
+    @hoard = true
+  end
+
+  def is_hoarded?
+    @hoard
```
aziobrow: great!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
+class ShoppingCart
+  attr_reader :name, :capacity, :products, :total_number_of_products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.delete('items')
```
aziobrow: sooo close! then it just needs to also be an integer, so this needed to have a `to_i` on the end here

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+    @products = []
+    @total_number_of_products = 0
```
aziobrow: this doesn't need to be an instance variable-- since you have a method whose job it is to do this work, anywhere you call the instance variable, you could just be calling the method.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def add_product(product)
+    @products << product
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def total_number_of_products
+    @products.each do |product|
+      @total_number_of_products += product.quantity.to_i
+    end
+    @total_number_of_products
```
aziobrow: Good! If you were just to use a local variable here, it would look like this: 

  def total_number_of_products
    total = 0
    @products.each do |product|
      total += product.quantity.to_i
    end
    total
end 
```

Comment (aziobrow) [reply lib/shopping_cart.rb]: Ruby also has a beautiful enumerable that's built for exactly this job called `sum`

```
def total_number_of_products
    products.sum do |product|
        product.quantity
    end 
end
```

File: lib/shopping_cart.rb [hunk 10]
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def is_full?
+    if total_number_of_products >= @capacity.to_i
+      true
+    else
+      false
+    end
```
aziobrow: Great! As an additional consideration, this can technically be accomplished in a single line.

```
def is_full?
    total_number_of_products >= @capacity.to_i
end 
```

Since that statement returns a boolean in and of itself, you don't technically need the if/else. if that statement is true, the whole method will return true, and vice versa.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def products_by_category(category)
+    products_by_category = products.group_by { |product| product.category }
+    products_by_category[category]
```
aziobrow: ooh! group_by! cool! This definitely works, and I love that you're pushing yourself to use some tools outside of your comfort zone.  

This use case is probably textbook for the use of `find_all` (get a subset of the original collection that meet a certain criterion). So a more efficient alternative would be:

```
end 
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def details
+    details = {}
+    details[:name] = name
+    details[:capacity] = capacity.to_i
+    return details
```
aziobrow: beautiful! you don't need this explicit return statement here `details` would be fine (as opposed to `return details`)

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def percentage_occupied
+    ((total_number_of_products / capacity.to_f) * 100).round(2)
```
aziobrow: perfect.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def sorted_products_by_quantity
+    sorted_products = @products.sort_by { |product| product.quantity.to_i }
+    sorted_products.reverse
```
aziobrow: This is perfect! Fantastic job, and in my opinion, the best solution to this problem.  If you wanted to, you can do this as a one-liner...

`@products.sort_by { |product| product.quantity.to_i }.reverse`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def product_breakdown
+    product_breakdown = {}
+    @products.each do |product|
+      product_breakdown[product.category] = products_by_category(product.category)
+    end
+    product_breakdown
```
aziobrow: soooo great! this is some tricky stuff, and this looks fantastic. great work. nice use of your `products_by_category` method to help out.

test/product_test.rb
```diff
@@ -9,35 +9,40 @@ def test_it_exists
...
-    skip
     product = Product.new(:paper, 'toilet paper', 3.70, '10')
     assert_equal 3.70, product.unit_price
   end
 
   def test_it_has_a_quantity
-    skip
     product = Product.new(:paper, 'toilet paper', 3.70, '10')
-    assert_equal 10, product.quantity
+    assert_equal '10', product.quantity
```
aziobrow: this is cheating 🙂 The test was set up with what information you were being given-- your job was to handle it in code. You shouldn't be changing your tests unless you've actually made a mistake in your test-- in this case, it was asking you to change this quantity from a string to an integer in the class.

test/product_test.rb
```diff
@@ -9,35 +9,40 @@ def test_it_exists
...
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_can_hoard
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+    assert_equal false, product.is_hoarded?
+
+    product.hoard
+    assert_equal 30, product.quantity
```
aziobrow: you don't need to test this here, although it doesn't hurt anything.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,121 @@
...
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    assert_instance_of ShoppingCart, cart
+  end
+
+  def test_it_has_attributes
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    assert_equal 'King Soopers', cart.name
+    assert_equal '30', cart.capacity
```
aziobrow: not quite-- the interaction pattern tells us that this should be an integer.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,121 @@
...
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+
+    assert_equal false, cart.is_full?
+
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product4)
+
+    assert_equal true, cart.is_full?
```
aziobrow: love the progression testing here-- ensure dynamic behavior.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,121 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    cart.add_product(product4)
+
+    assert_equal [product1, product3], cart.products_by_category(:paper)
```
aziobrow: you could test for another category here too, to be sure it's dynamic and works for different types of categories

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  def test_it_can_return_cart_details
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    assert_equal cart.details, {:name=> "King Soopers", :capacity=> 30}
```
aziobrow: this order is flipped-- it's expected first, then actual. 

`assert_equal ({:name=> "King Soopers", :capacity=> 30}), cart.details`

We'll talk tomorrow about why you need parentheses around the hash here

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,121 @@
...
+    tissue_paper = Product.new(:paper, 'tissue paper', 1.25, '1')
+    apples = Product.new(:produce, 'apples', 0.99, '20')
+    chicken = Product.new(:meat, 'chicken', 4.50, '2')
+    cart.add_product(toilet_paper)
+    cart.add_product(tissue_paper)
+    cart.add_product(apples)
+    cart.add_product(chicken)
+
+    assert_equal cart.product_breakdown, {:paper => [toilet_paper, tissue_paper], :meat => [chicken], :produce => [apples]}
+  end
```
aziobrow: these are great-- i assumed you named those variables differently because they helped you conceptualize, which is totally fine. 

i'd say as a general process though, we tend to name things their class- that way future you/other developers can easily see that those are instances of the product class, and not, let's say, instances of the apple class or chicken class, etc.

lib/product.rb
stellakunzang: that makes a lot more sense than calling `quantity.to_i` EVERYWHERE

lib/product.rb
stellakunzang: I misunderstood - I thought we were supposed to get creative about what we thought `hoard` would do

lib/shopping_cart.rb
stellakunzang: yes! Just like the problem with quantity...I'm seeing a pattern here.

lib/shopping_cart.rb
stellakunzang: This makes sense. I think I need to implement it more to drill it into my head that this will return the boolean that I want.

test/shopping_cart_test.rb
stellakunzang: oh good! this was really tripping me up

PR #18, opened 2020-03-25
Michael Gallup

Gallup93:


timomitchel:
Nice job committing throughout the challenge!

lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_f
+  end
+    @hoarded = false
```
timomitchel: This `@hoarded` instance variable has class scope if you put it here. We want our instance variables created within the initialize method.

lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
+  attr_reader :category, :name, :unit_price, :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_f
```
timomitchel: We actually wanted this to be an integer, so `.to_i` was the way to accomplish that here.

lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
+  end
+    @hoarded = false
   end
 
   def total_price
     @unit_price * @quantity
+
+  def hoard
```
timomitchel: These two methods `hoard` and `is_hoarded?` are set up the way we were hoping, but because of where you originally set the `@hoarded` instance variable, they don't pass our tests.

lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_f
+  end
+    @hoarded = false
   end
 
   def total_price
     @unit_price * @quantity
```
timomitchel: You're missing an `end` for `total_price`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,51 @@
...
+    @products = []
+    @is_full = false
+    @details = {name: @name, capacity: @capacity}
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
```
timomitchel: Nice job using `each` here to get the job done. You could save yourself some work by using the `sum` enum. 
```
@products.sum do |product|
      product.quantity
    end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,51 @@
...
+    number = 0
+    @products.each do |product|
+      number += product.quantity
+    end
+    number
+  end
+
+  def is_full?
+    if total_number_of_products >= @capacity
```
timomitchel: No need for this `if` statement. We only need the condition. But, attention to detail here, because the interaction pattern said that a shopping cart is full if it has more products in it than its capacity. So we wanted:

lib/shopping_cart.rb
```diff
@@ -0,0 +1,51 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def details
```
timomitchel: This could be added as an `attr_reader` to accomplish the same functionality.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,51 @@
+class ShoppingCart
+  attr_reader :name, :capacity, :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity[0..1].to_i
+    @products = []
+    @is_full = false
+    @details = {name: @name, capacity: @capacity}
```
timomitchel: Nice job building this hash here.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    @products.each do |product|
+      percent += product.quantity
+    end
+
+    ((percent / @capacity) * 100).round(2)
+  end
+
+  def sorted_products_by_quantity
+    @products.sort_by {|product| product.quantity}
```
timomitchel: Attention to detail here, this gives us the products in ascending order, but the interaction pattern called for descending, which could be accomplished by throwing a `.reverse` on the end of this block:
`@products.sort_by {|product| product.quantity}.reverse`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,51 @@
...
+      false
+    end
+  end
+
+  def details
+    @details
+  end
+
+  def percentage_occupied
+    percent = 0
```
timomitchel: You duplicate the work you did in `total_number_of_products`. You could have called that method in place of creating the `percent` variable:
`((total_number_of_products / @capacity) * 100).round(2)`

test/product_test.rb
```diff
@@ -9,35 +9,46 @@ def test_it_exists
...
+    # skip
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_if_is_hoarded?
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
```
timomitchel: Nice job with the test setup an execution here.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/product'
+require './lib/shopping_cart'
+
+class ShoppingCartTest < Minitest::Test
+
+  def test_initialization
```
timomitchel: This is a solid attributes test, but you aren't testing `initialization`, you're testing that you can set and query attributes correctly. We'd also like to see a `test_it_exists`  that uses `assert_instance_of` as the first test, because this will ensure proper TDD.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    cart.add_product(product1)
+    cart.add_product(product2)
+
+    assert_equal [product1, product2], cart.products
+  end
+
+  def test_total_number_of_products
+    cart = ShoppingCart.new("King Soopers", "30items")
```
timomitchel: Exactly what we want here. Nice job.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+
+    assert_equal false, cart.is_full?
```
timomitchel: Great job testing both sides of this boolean, unfortunately it didn't catch the misstep in the comparison: `>` instead of `>=`.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    cart.add_product(product4)
+
+    assert_equal true, cart.is_full?
+  end
+
+  def test_details
+    cart = ShoppingCart.new("King Soopers", "30items")
```
timomitchel: You could set up a local variable to hold your hash here, so that you can use the `assert_equal`:
```
expected = {name: "King Soopers", capacity: 30}
assert_equal expected, cart.details
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+
+    assert_equal 43.33, cart.percentage_occupied
+  end
+
+  def test_sorted_by_quantity
+    cart = ShoppingCart.new("King Soopers", "30items")
```
timomitchel: The way you built your test failed you here. Make sure to pay attention to detail in the interaction patter of which objects are returned where. You could have checked the object_id(s) in the interaction pattern to note that the array that is returned is in descending order.

PR #20, opened 2020-03-25
Eric Larson

EricLarson2020:
Shopping_cart_2003 week 2 challenge

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,10 +1,10 @@
-class Prodct
-
-  def initialize(category, name, unit_price, quantity)
+class Product
+  attr_reader :category, :name, :unit_price, :quantity
+  def initialize(category, name, unit_price = 3.70, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
   end
 
   def total_price
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -42,4 +42,9 @@ class ShoppingCart
     }
     details_hash
   end
+
+  def percentage_occupied
+
+    (total_number_of_products.to_f/capacity.to_f).round(4) * 100
+  end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -47,4 +47,41 @@ class ShoppingCart
 
     (total_number_of_products.to_f/capacity.to_f).round(4) * 100
   end
+
+  def sorted_products_by_quantity
+
+    highest_quantity = 0
+    sorted_quantity = []
+    products.each do |product|
+      if  product.quantity >= highest_quantity
+        highest_quantity = product.quantity
+        sorted_quantity.unshift(product)
+    elsif product.quantity < highest_quantity
+        sorted_quantity << product
+      end
+    end
+  sorted_quantity
+  end
+
+  def product_breakdown
+
+    paper = products.select do |product|
+      product.category == :paper
+      end
+
+    meat = products.select do |product|
+      product.category == :meat
+      end
+
+    produce = products.select do |product|
+      product.category == :produce
+    end
+
+    hash_breakdown = {
+      :meat => meat,
+      :paper => paper,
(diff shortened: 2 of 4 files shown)


lib/product.rb
```diff
@@ -1,10 +1,10 @@
-class Prodct
-
-  def initialize(category, name, unit_price, quantity)
+class Product
+  attr_reader :category, :name, :unit_price, :quantity
+  def initialize(category, name, unit_price = 3.70, quantity)
```
memcmahon: This definitely works, but we don't really need to give unit_price a default value.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,87 @@
+require './lib/product'
```
memcmahon: Since we don't create any products in this class, we don't need to require in this file.  I would only require files when you see that class being called directly in a file (`Product.new(...)` for example)

lib/shopping_cart.rb
```diff
@@ -0,0 +1,87 @@
...
+    @products << product
+  end
+
+  def total_number_of_products
+
+    total_quantity = 0
+    products.each do |product|
+      total_quantity += product.quantity
+    end
+    total_quantity
```
memcmahon: Great use of `each`!  You could also use the enumerable `sum`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def is_full?
+    total_number_of_products > @capacity
+  end
+
+  def products_by_category(category)
+
+
+    catagory_products= products.select do |product|
```
memcmahon: With `select` you do not need to assign this variable; you can rely on the return value of the select statement itself.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,87 @@
...
+    catagory_products
+  end
+
+  def details
+
+    details_hash = {
+      :name => name,
+      :capacity => capacity
+    }
+    details_hash
```
memcmahon: This works great, but has a little more 'stuff' than is strictly necessary - you could do something like this:

```ruby
def details
  {
    name: @name,
    capacity: @capacity
  }
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,87 @@
...
+    products.each do |product|
+      if  product.quantity >= highest_quantity
+        highest_quantity = product.quantity
+        sorted_quantity.unshift(product)
+    elsif product.quantity < highest_quantity
+        sorted_quantity << product
+      end
+    end
+  sorted_quantity
+  end
```
memcmahon: Great job working through this sort logic!  You could also use the enumerable `sort_by`: 

```ruby
@products.each do |product|
  product.quantity
end.reverse
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,87 @@
...
+    produce = products.select do |product|
+      product.category == :produce
+    end
+
+    hash_breakdown = {
+      :meat => meat,
+      :paper => paper,
+      :produce => produce
+    }
+    hash_breakdown
```
memcmahon: This will work for these three categories, but not for any number of categories.  Either of the two options below would work better:

```ruby
  product.category
end
```

```ruby
breakdown = Hash.new
@products.each do |product|
  breakdown[product.category] = [] if breakdown[product.category].nil?
  breakdown[product.category] << product
end
breakdown
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  def test_the_cart_has_products_by_category
+    cart = ShoppingCart.new("King Soopers", "30items")
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    assert_equal false, cart.is_full?
```
memcmahon: You don't need to make this assertion again - it is tested above and can be removed from other tests.

PR #21, opened 2020-03-25
Derek Borski

dborski:
Here is my shopping_cart_2003 independent challenge!

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -10,6 +10,7 @@ class Product
     @name = name
     @unit_price = unit_price
     @quantity = quantity.to_i
+    @is_hoarded = false
   end
 
   def total_price
@@ -17,6 +18,10 @@ class Product
   end
 
   def is_hoarded?
-    false
+    @is_hoarded
+  end
+
+  def hoard
+    @is_hoarded = true
   end
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,8 +1,9 @@
-class Prodct
+class Product
 
+  attr_reader :category, :name
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
+    @name = name
     @unit_price = 3.70
     @quantity = quantity
   end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -4,7 +4,6 @@ class Product
               :name,
               :unit_price,
               :quantity
-
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -15,4 +15,8 @@ class Product
   def total_price
     @unit_price * @quantity
   end
+
+  def is_hoarded?
+    false
+  end
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,11 +1,15 @@
 class Product
 
-  attr_reader :category, :name
+  attr_reader :category,
+              :name,
+              :unit_price,
+              :quantity
+
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
-    @unit_price = 3.70
-    @quantity = quantity
+    @unit_price = unit_price
+    @quantity = quantity.to_i
   end
 
   def total_price
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,0 +1,12 @@
+class ShoppingCart
+
+attr_reader :name, :capacity
+def initialize(name, capacity)
+  @name = name
+  @capacity = capacity.to_i
+end
+
+
+
+
+end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -2,7 +2,8 @@ class ShoppingCart
 
   attr_reader :name,
               :capacity,
-              :products
+              :products,
+              :categories
   def initialize(name, capacity)
     @name = name
     @capacity = capacity.to_i
@@ -23,6 +24,13 @@ class ShoppingCart
 
   def is_full?
     total_number_of_products > 30
+  end
+
+  def products_by_category(category)
+    sorted_categories = products.find_all do |product|
+       product.category == category
+    end
+    sorted_categories
   end
 
 
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -37,4 +37,9 @@ class ShoppingCart
     cart_details = Hash[:name, name, :capacity, capacity]
     cart_details
   end
+
+  def percentage_occupied
+    percentage_occupied = (total_number_of_products.to_f / capacity.to_f) * 100
+    percentage_occupied.round(2)
+  end
 end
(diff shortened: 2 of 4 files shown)


timomitchel:
Nice job committing throughout the challenge.

lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
+              :unit_price,
+              :quantity
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
```

lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
+    @is_hoarded = false
```
timomitchel: Really nice job coming up with an instance variable of your own to store this value.

lib/product.rb
```diff
@@ -1,13 +1,26 @@
...
   end
 
   def total_price
     @unit_price * @quantity
   end
+
+  def is_hoarded?
```
timomitchel: Exactly what we were hoping to see.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,45 @@
...
+    @name = name
+    @capacity = capacity.to_i
+    @products = []
+  end
+
+  def add_product(product)
+    products << product
+  end
+
+  def total_number_of_products
```
timomitchel: Great job getting each to get the job done here. The `sum` enum can save you a few lines of code:

lib/shopping_cart.rb
```diff
@@ -0,0 +1,45 @@
...
+    sum_of_products = 0
+    products.each do |product|
+      sum_of_products += product.quantity
+    end
+    sum_of_products
+  end
+
+  def is_full?
+    total_number_of_products > capacity
```
timomitchel: Great attention to detail here. This is what the interaction pattern called for.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def products_by_category(category)
+    sorted_categories = products.find_all do |product|
```
timomitchel: Great job choosing the correct enum for the job here! You could simplify this to:
```
@products.find_all do |product|
product.category == category
end
```
which will provide the correct return value and avoid the variable assignment.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,45 @@
...
+       product.category == category
+    end
+    sorted_categories
+  end
+
+  def details
+    cart_details = Hash[:name, name, :capacity, capacity]
```
timomitchel: This is an interesting way to create a hash, but it breaks ruby convention. We'd rather see: 
`{name: @name, capacity: @capacity}`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,45 @@
...
+    cart_details
+  end
+
+  def percentage_occupied
```
timomitchel: Clean and readable.

test/product_test.rb
```diff
@@ -9,35 +9,41 @@ def test_it_exists
...
   def test_it_can_get_total_price
-    skip
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
-
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_is_hoarded_is_false_be_default
```
timomitchel: Nice job testing both sides of the boolean condition here. Because you do this in `test_hoard_product_changes_is_hoarded_to_true` there is no need for `test_is_hoarded_is_false_be_default`.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+require './lib/product'
+require './lib/shopping_cart'
+
+class ShoppingCartTest < Minitest::Test
+  def test_it_exists
+    cart = ShoppingCart.new("King Soopers", "30items")
+    assert_instance_of ShoppingCart, cart
+  end
+
+  def test_it_has_a_name
```
timomitchel: All of the attributes could be combined into one attributes test.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+  def test_can_add_products_to_cart
+    cart = ShoppingCart.new("King Soopers", "30items")
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    assert_equal [product1, product2], cart.products
+  end
+
+  def test_total_number_of_products_in_cart
```
timomitchel: Great setup and execution throughout this test and the previous test.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    assert_equal 13, cart.total_number_of_products
+  end
+
+  def test_cart_is_not_full_by_default
```
timomitchel: This test and the next test could be combined into one `is_full?` test.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    cart.add_product(product4)
+    assert_equal true, cart.is_full?
+  end
+
+  def test_cart_can_be_sorted_by_category
```
timomitchel: Great job robustly testing all categories here!

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    cart.add_product(product3)
+    cart.add_product(product4)
+    assert_equal [product1, product3], cart.products_by_category(:paper)
+    assert_equal [product2], cart.products_by_category(:meat)
+    assert_equal [product4], cart.products_by_category(:produce)
+  end
+
+  def test_details_of_cart_in_hash
+    cart = ShoppingCart.new("King Soopers", "30items")
+    assert_equal cart_details = {:name=> "King Soopers", :capacity=> 30}, cart.details
```
timomitchel: We don't want to set the variable on the same line as the `assert_equal`. We want to move: 
`cart_details = {:name=> "King Soopers", :capacity=> 30}`
one line above our `assert_equal` and then call:
`assert_equal cart_details, cart.details`

PR #23, opened 2020-03-25
Shopping cart

JBelta:


JBelta:
James Belta
Shopping Cart, 

I have wasted so much time looking for ";" and "end" when an error pops up. Gonna look for atom packages to watch over proper syntax!

lib/product.rb
```diff
@@ -1,13 +1,43 @@
...
+    @name
+  end
+
+  def unit_price
+    @unit_price
+  end
+
+  def quantity
+    @quantity.to_i
+  end
```
memcmahon: These _could_ be attr_readers, with one small change to quantity:

```ruby
class Product
  attr_reader :category, :name, :unit_price, :quantity
  
    @category = category
    @name = name
    @unit_price = unit_price
    @quantity = quantity.to_i
  end
end
```

lib/product.rb
```diff
@@ -1,13 +1,43 @@
...
+      false
+    elsif @quantity.to_i == 0
+      true
+    end
   end
+
+  def hoard
+    @quantity =
+    (@quantity.to_i - @quantity.to_i).to_s
+  end
```
memcmahon: While this logic might make sense, it is not what was outlined in the interaction pattern (which did not make any mention of a specific amount indicating hoard-ness).  What we were looking for was:

```ruby
class Product
  ...
    ...
    @is_hoarded = false
  end

  def is_hoarded?
    @is_hoarded
  end
  
  def hoard
    @is_hoarded = true
  end
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,32 @@
...
+    @name
+  end
+
+  def capacity
+    @capacity[0..1].to_i
+  end
+
+  def products
+    @products
+  end
```
memcmahon: Similar to the Product, these _could_ be attr_readers.  They don't have to be, but you will generally see rubyists leveraging attrs whenever possible.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,32 @@
...
+  def add_product(item)
+    @products << item
+  end
+
+#on to iter 3
+
+def is_full?
+  if @products.quantity > @capacity.capacity
+    false
+
```
memcmahon: Try not to turn things in with missing ends!!

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/product'
+require './lib/shopping_cart'
+
+class ShoppingCartTest < Minitest::Test
```
memcmahon: These tests look great, though I am surprised to not see a test for the `is_full?` method because you do have some code for that method - remember to write tests first.

PR #24, opened 2020-03-25
Antonio Jackson

AntonioJacksonII:
Independent Challenge #2 for Antonio Jackson

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,8 +1,8 @@
-class Prodct
+class Product
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
+    @name = name
     @unit_price = 3.70
     @quantity = quantity
   end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -7,9 +7,18 @@ class Product
     @name = name
     @unit_price = unit_price
     @quantity = quantity.to_i
+    @is_hoarded = false
   end
 
   def total_price
     @unit_price * @quantity
+  end
+
+  def is_hoarded?
+    @is_hoarded
+  end
+
+  def hoard
+    @is_hoarded = true
   end
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -5,7 +5,7 @@ class Product
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
-    @unit_price = 3.70
+    @unit_price = unit_price
     @quantity = quantity.to_i
   end
 
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,5 +1,7 @@
 class Product
 
+  attr_reader :category, :name, :unit_price, :quanitty
+
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,12 +1,12 @@
 class Product
 
-  attr_reader :category, :name, :unit_price, :quanitty
+  attr_reader :category, :name, :unit_price, :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
     @unit_price = 3.70
-    @quantity = quantity
+    @quantity = quantity.to_i
   end
 
   def total_price
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,0 +1,9 @@
+class ShoppingCart
+
+  attr_reader :name, :capacity
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity
+  end
+end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -24,4 +24,11 @@ class ShoppingCart
     total_number_of_products > @capacity
   end
 
+  def products_by_category(category)
+    @products.select do |product|
+      product.category == category
+    end
+  end
+
+
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -38,5 +38,9 @@ class ShoppingCart
     ((total_number_of_products.to_f / @capacity) * 100).round(2)
   end
 
+  def sorted_products_by_quantity
+    @products.sort_by(&:quantity).reverse
+  end
+
 
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -42,5 +42,7 @@ class ShoppingCart
     @products.sort_by(&:quantity).reverse
   end
 
-
+  def product_breakdown
+    @products.sort_by(&:category)
+  end
 end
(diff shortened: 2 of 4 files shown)


lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
+    @is_hoarded = false
```
aziobrow: great job with these

aziobrow: especially nice job catching that quantity needed to be an integer

lib/product.rb
```diff
@@ -1,13 +1,24 @@
...
     @unit_price * @quantity
   end
+
+  def is_hoarded?
+    @is_hoarded
+  end
+
+  def hoard
+    @is_hoarded = true
+  end
```
aziobrow: perfect!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
+class ShoppingCart
+
+  attr_reader :name, :capacity, :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.to_i
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
...
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    total_products = 0
+     @products.each do |product|
+       total_products += product.quantity
+     end
+     total_products
```
aziobrow: this is great! Ruby also has a perfect method for this that's designed to do exactly this job:

```
def total_number_of_products
    products.sum do |product|
        product.quantity
    end 
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def is_full?
+    total_number_of_products > @capacity
```
aziobrow: really nicely done-- love that you accomplished this in a single line.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def products_by_category(category)
+    @products.select do |product|
```
aziobrow: fantastic use of select here. this is exactly the right use case, and you executed it perfectly.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
...
+      product.category == category
+    end
+  end
+
+  def details
+  {:name => @name, :capacity => @capacity}
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def percentage_occupied
+    ((total_number_of_products.to_f / @capacity) * 100).round(2)
```
aziobrow: perfect

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def sorted_products_by_quantity
+    @products.sort_by(&:quantity).reverse
```
aziobrow: nice! even using shorthand syntax... great work.  this is the way i would have solved this also!

lib/shopping_cart.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def product_breakdown
+    @products.sort_by(&:category)
```
aziobrow: this doesn't quite work-- this gives you an array of products sorted by their categories (which I assume will default to alphabetical sorting).  You may have run out of time in the middle of this one, but this wouldn't pass your test.  Something like this would work for the time being (I'd use a tool called `reduce`, but we'll save that for a bit later in the mod):

```
def product_breakdown
     product_breakdown = {}
     @products.each do |product|
          product_breakdown[product.category] = products_by_category(product.category)
      end
    product_breakdown
end
```

test/product_test.rb
```diff
@@ -9,35 +9,39 @@ def test_it_exists
...
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_can_be_hoarded
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+
+    assert_equal false, product.is_hoarded?
+
+    product.hoard
+    assert_equal true, product.is_hoarded?
```
aziobrow: love the progression testing here to ensure dynamic functionality

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,139 @@
...
+  def test_it_has_a_capacity
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    assert_equal 30, cart.capacity
+  end
+
+  def test_it_can_hold_products
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    assert_equal [], cart.products
```
aziobrow: great! you can combine the attributes into one test too, if you'd like.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,139 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product4)
+
+    assert_equal [product1, product3], cart.products_by_category(:paper)
```
aziobrow: you could test for another category here too, to ensure it doesn't just work for paper products.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,139 @@
...
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+    cart.add_product(product4)
+
+    assert_equal ({:meat => [product2], :paper => [product1, product3], :produce => [product4]}), cart.product_breakdown
```
aziobrow: nice job figuring out that you needed to use parentheses here to make it an assertion.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,139 @@
...
+  end
+end
```
aziobrow: All of your tests look great! You did a fantastic job of sticking exactly to the interaction pattern, and thus your implementation is exactly correct for everything but the last method 😁

PR #25, opened 2020-03-25
Eric Hale

EHale64:
I totally forgot to make commits as I was going!

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,13 +1,22 @@
-class Prodct
+class Product
+  attr_reader :category, :name, :unit_price, :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
   end
 
+  def quantity
+    @quantity.to_i
+  end
+
   def total_price
-    @unit_price * @quantity
+    @unit_price * quantity
+  end
+
+  def is_hoarded?
+    false
   end
 end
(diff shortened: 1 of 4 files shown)


lib/product.rb
```diff
@@ -1,13 +1,22 @@
...
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
+    @name = name
+    @unit_price = unit_price
     @quantity = quantity
   end
 
+  def quantity
```
timomitchel: We never want an `attr_reader` and a method that are named the same and refer to the same instance variable. You could accomplish this behavior by calling `.to_i` on the `quantity` parameter within the initialize method. Or, you could remove the `attr_reader` for `quantity`.
```
@quantity = quantity.to_i
```

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    cart = ShoppingCart.new("King Soopers", "30items")
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    cart.add_product(product1)
+    cart.add_product(product2)
+
+    assert_equal 12, cart.total_number_of_products
+  end
+
+  def test_it_can_be_full
```
timomitchel: Whenever we have a method that returns a boolean we want to make sure we test both the scenario that makes the method return a `true` and the scenario that makes it return a false. So, a: 
```
assert_equal false,  cart.is_full?
```
between lines 59 & 60, would make this test more robust, and thus would follow the interaction pattern exactly.

PR #27, opened 2020-03-25
Kyle Iverson

kiverso:
Fix typos and bugs for product class and added hoard functionality. Added shopping cart class and tests.

### DIFF
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,13 +1,14 @@
 class ShoppingCart
-  attr_reader :name, :capacity, :products
+  attr_reader :name, :capacity, :products, :total_number_of_products
   def initialize(name, capacity, products = [])
     @name = name
     @capacity = capacity.to_i
     @products = products
+    @total_number_of_products = 0
   end
 
   def add_product(product)
     @products << product
+    @total_number_of_products += product.quantity
   end
-
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -11,4 +11,8 @@ class ShoppingCart
     @products << product
     @total_number_of_products += product.quantity
   end
+
+  def is_full?
+    @total_number_of_products > @capacity
+  end
 end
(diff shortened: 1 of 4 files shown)


lib/shopping_cart.rb
```diff
@@ -0,0 +1,30 @@
+class ShoppingCart
+  attr_reader :name, :capacity, :products, :total_number_of_products
+  def initialize(name, capacity, products = [])
+    @name = name
+    @capacity = capacity.to_i
+    @products = products
+    @total_number_of_products = 0
```
memcmahon: Having a @total_number_of_products instance variable, that we update when we add products works, but we generally want to reduce the number of instance variables that exists in any class.  Since the `@products` array already contains the information we need to calculate the `total_number`, we can use that single storage of data to calculate what we want on the fly, rather than duplicating the storage of that information.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,30 @@
...
+    @products << product
+    @total_number_of_products += product.quantity
+  end
+
+  def is_full?
+    @total_number_of_products > @capacity
+  end
+
+  def products_by_category(category)
+    products.select {|product| product.category == category}
```
memcmahon: Great use of select!

test/product_test.rb
```diff
@@ -9,35 +9,40 @@ def test_it_exists
...
   def test_it_can_get_total_price
-    skip
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_can_hoard
```
memcmahon: This test looks great!

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    cart.add_product(product4)
+
+    assert_equal [product1, product3], cart.products_by_category(:paper)
+    assert_equal [product2], cart.products_by_category(:meat)
+  end
+
+  def test_it_can_get_details
+    cart = ShoppingCart.new("King Soopers", "30items")
+
+    details_for_test = {:name=> "King Soopers", :capacity=> 30}
```
memcmahon: Good job figuring out how to make an assertion on a hash!  It is a bit more conventional to call this variable `expected`

PR #28, opened 2020-03-25
Sage Lee Shopping Cart Ind Challenge

sagemlee:
Added tests and file updates for products and shopping cart.

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -7,9 +7,19 @@ class Product
     @name = name
     @unit_price = unit_price
     @quantity = quantity.to_i
+    @hoarded = false
   end
 
   def total_price
     @unit_price * @quantity
   end
+
+  def is_hoarded?
+    @hoarded
+  end
+
+  def hoard
+    @hoarded = true
+  end 
+
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -20,6 +20,6 @@ class Product
 
   def hoard
     @hoarded = true
-  end 
+  end
 
 end
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,10 +1,12 @@
-class Prodct
+class Product
+    attr_reader :category, :name, :quantity
+    attr_accessor :unit_price
 
-  def initialize(category, name, unit_price, quantity)
+  def initialize(category, name, unit_price = 3.70, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
   end
 
   def total_price
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -6,11 +6,39 @@ class ShoppingCart
 
   def initialize(name, capacity)
     @name = name
-    @capacity = capacity
+    @capacity = capacity.to_i
     @products = []
   end
 
   def add_product(new_product)
     @products << new_product
   end
+
+  def total_number_of_products
+    total = 0
+      @products.each do |item|
+      total += item.quantity
+      end
+      return total
+
+    end
+
+    def is_full?
+     self.total_number_of_products >= @capacity
+    end
+
+    def products_by_category(searched_category)
+      category_set = []
+      @products.each do |item|
+        if item.category == searched_category
+          category_set << item
+        end
+      end
+      return category_set
+    end
+
+    def details
+      details_hash[:name] = @name
+      details_hash[:capacity] = @capacity 
+    end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -1,0 +1,16 @@
+require './lib/product'
+require'pry'
+
+class ShoppingCart
+  attr_reader :name, :capacity, :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity
+    @products = []
+  end
+
+  def add_product(new_product)
+    @products << new_product
+  end
+end
--- a/test/product_test.rb
+++ b/test/product_test.rb
@@ -40,4 +40,15 @@ class ProductTest < Minitest::Test
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_starts_unhoarded
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+    refute product.is_hoarded?
+  end
+
+  def test_it_can_be_hoarded
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product.hoard
+    assert product.is_hoarded?
+  end
 end
--- a/test/product_test.rb
+++ b/test/product_test.rb
@@ -1,6 +1,7 @@
 require 'minitest/autorun'
 require 'minitest/pride'
 require './lib/product'
+require'pry'
 
 class ProductTest < Minitest::Test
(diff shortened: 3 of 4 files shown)


lib/product.rb
```diff
@@ -1,13 +1,25 @@
-class Prodct
+class Product
+    attr_reader :category, :name, :quantity
+    attr_accessor :unit_price
```
timomitchel: No need for this accessor. `unit_price` should be an `attr_reader`

lib/product.rb
```diff
@@ -1,13 +1,25 @@
...
 
-  def initialize(category, name, unit_price, quantity)
+  def initialize(category, name, unit_price = 3.70, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i
```

lib/product.rb
```diff
@@ -1,13 +1,25 @@
...
+    @hoarded = false
```
timomitchel: Really nice job coming up with an instance variable of your own to store this value.

lib/product.rb
```diff
@@ -1,13 +1,25 @@
...
   end
 
   def total_price
     @unit_price * @quantity
   end
+
+  def is_hoarded?
+    @hoarded
```
timomitchel: Exactly what we were hoping to see.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,44 @@
+require './lib/product'
```
timomitchel: No need to require this file, because we never explicitly refer to the `Product` class in this class. We pass in, hold onto, and manipulate `Product` objects, but we instantiate them in the test, so `ShoppingCart` doesn't need to know about this file, because when it gets passed objects, they are already fully formed. Only if we called `Product.new(..)` in this file, would we need to require the product file.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,44 @@
...
+    @capacity = capacity.to_i
+    @products = []
+  end
+
+  def add_product(new_product)
+    @products << new_product
+  end
+
+  def total_number_of_products
+    total = 0
```
timomitchel: Great job getting each to get the job done here. The `sum` enum can save you a few lines of code:

lib/shopping_cart.rb
```diff
@@ -0,0 +1,44 @@
...
+      @products.each do |item|
+      total += item.quantity
+      end
+      return total
```
timomitchel: No need for this explicit `return` here. The final line of the method will always be what is returned.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    end
+
+    def is_full?
+     self.total_number_of_products >= @capacity
```
timomitchel: Remember to pay attention to detail. Based on the exact wording of the interaction pattern, this should be `>` not `>=`.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def products_by_category(searched_category)
+      category_set = []
+      @products.each do |item|
```
timomitchel: Great job using each to get the job done here. `find_all` would save you a few lines here: 
`@products.find_all { |item| item.category == searched_category}`

lib/shopping_cart.rb
```diff
@@ -0,0 +1,44 @@
...
+        if item.category == searched_category
+          category_set << item
+        end
+      end
+      return category_set
```
timomitchel: Again, no need for the explicit return here.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def details
```
timomitchel: This would have worked well if you would have defined the hash first, and returned it at the end of the method;
```
def details
      details_hash = {}
      details_hash[:name] = @name
      details_hash[:capacity] = @capacity
      details_hash
    end
```
But, remember, we don't want to include the datatype within our naming of variables or methods.

test/product_test.rb
```diff
@@ -9,35 +10,45 @@ def test_it_exists
...
-    skip
+
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
-
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_starts_unhoarded
```
timomitchel: Nice job testing both possible booleans of the `is_hoarded` method. Be careful with `assert` and `refute` as they will show a passing test for any truthy (assert) or falsy (refute) values.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+ def test_it_has_readable_attributes
+   cart = ShoppingCart.new("King Soopers", "30items")
+
+   assert_equal "King Soopers", cart.name
+   assert_equal 30, cart.capacity
+ end
+
+ def test_it_starts_with_no_products
+   cart = ShoppingCart.new("King Soopers", "30items")
+   assert_equal [], cart.products
```
timomitchel: This `assert_equal` could be moved into the attribute test above.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+   cart.add_product(product1)
+   cart.add_product(product2)
+
+   assert_equal [product1, product2], cart.products
+ end
+
+
+ def test_it_can_total_product_amounts
+   cart = ShoppingCart.new("King Soopers", "30items")
+   product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
```
timomitchel: Great test setup and execution here.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+
+    refute cart.is_full?
+  end
+
+  def test_it_can_be_full
```
timomitchel: Again, nice job testing both sides of the boolean condition. You could combine these `is_full?` tests into one to save yourself some time.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
+    product4 = Product.new(:produce, 'apples', 0.99, '20')
+
+    cart.add_product(product1)
+    cart.add_product(product2)
+    cart.add_product(product3)
+    cart.add_product(product4)
+
+    assert_equal [product1, product3], cart.products_by_category(:paper)
```
timomitchel: You should test all of the categories for this method to make the testing more robust.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  def test_it_details_hash
+    assert_equal {:name=> "King Soopers", :capacity=> 30}, cart.details
```
timomitchel: I can see where this probably caused you some issues. First off, `cart` is nil currently, and we either need to pass a hash in parentheses to make minitest happy, or we could save the hash to a local variable and then pass the `assert_equal` the local variable.

PR #29, opened 2020-03-25
Judith Pillado 

judithpillado:


timomitchel:
Nice job making commits throughout!

lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
     @category = category
     @name = name_parameter
-    @unit_price = 3.70
+    @unit_price = unit_price
     @quantity = quantity
   end
 
   def total_price
-    @unit_price * @quantity
+    @unit_price * (@quantity.to_f)
```
timomitchel: I like that you changed the datatype of `@quantity` to match the datatype of `@unit_price`. It's always best practice to ensure we are working with the same datatypes before we do a calculation.

lib/product.rb
```diff
@@ -1,13 +1,23 @@
...
   end
+
+  def is_hoarded?
```
timomitchel: This method would be best served to return an instance variable that holds a boolean. Based on the interaction pattern, it would start out as `false`. If you add a `@hoarded` instance variable to initialize with an initial value of false, then based on the interaction pattern we could have this method read: 
```
def is_hoarded?
@hoarded
end
```

Then, our `hoard` method would change the value of `@hoarded` to true when it was called like this: 
```
def hoard
@hoarded = true
end
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,22 @@
+class ShoppingCart
+  attr_reader :name, :capacity
+  attr_accessor :products
```
timomitchel: No need for an accessor here. We only need an `attr_reader` for each attribute.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize(name_parameter, capacity)
+    @name = name_parameter
+    @capacity = capacity
+    @products = []
+  end
+
+  def capacity
```
timomitchel: We don't want a method and an `attr_reader` with the same name. We only want one of these to return the value of our instance variable, but not both. We could also change the datatype of capacity in initialize on line 7 to avoid needing to define this method:
```
@capacity = capacity.to_i
```

test/product_test.rb
```diff
@@ -9,35 +9,36 @@ def test_it_exists
...
-
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_it_is_hoarded?
+    product = Product.new(:paper, 'toilet paper', 3.70, '10')
+    assert_equal false, product.is_hoarded?
+    product.hoard
+    assert_equal true, product.is_hoarded
```
timomitchel: This test setup looks good other than this `is_hoarded` missing a `?`. This probably gave you an undefined method error here.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,33 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/product'
+require './lib/shopping_cart'
+
+class ShoppingCartTest < Minitest::Test
+  def test_it_exists
```
timomitchel: Your test setup and execution in this file look good.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    cart.add_product(product1)
+    cart.add_product(product2)
+    assert_equal [product1, product2], cart.products
+  end
+
+  def test_total_number_of_products
+    cart = ShoppingCart.new("King Soopers", "30items")
+    product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
+    product2 = Product.new(:meat, 'chicken', 4.50, '2')
+    product3 = Product.new(:paper, 'tissue paper', 1.25, '1')
```
timomitchel: The final line of this test method should read:
```
assert_equal 13, cart.total_number_of_products
```

PR #30, opened 2020-03-25
Ashkan Abbasi

Ashkanthegreat:
Got stuck on last two methods in Iteration 3.  I think my method was correct for products by category, but kept getting nil value, so I guess not.

### DIFF
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,10 +1,11 @@
-class Prodct
+class Product
+  attr_reader :category, :name, :unit_price, :quantity
 
   def initialize(category, name, unit_price, quantity)
     @category = category
-    @name = name_parameter
-    @unit_price = 3.70
-    @quantity = quantity
+    @name = name
+    @unit_price = unit_price
+    @quantity = quantity.to_i 
   end
 
   def total_price
--- a/lib/product.rb
+++ b/lib/product.rb
@@ -1,14 +1,19 @@
 class Product
-  attr_reader :category, :name, :unit_price, :quantity
+  attr_reader :category, :name, :unit_price, :quantity, :hoarded
 
   def initialize(category, name, unit_price, quantity)
     @category = category
     @name = name
     @unit_price = unit_price
-    @quantity = quantity.to_i 
+    @quantity = quantity.to_i
+    @hoarded = false
   end
 
   def total_price
     @unit_price * @quantity
+  end
+
+  def hoard
+    @hoarded = true 
   end
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -19,4 +19,20 @@ class ShoppingCart
     @products.inject(0) {|result, product| result + product.quantity } > @capacity
   end
 
+  def products_by_category(category)
+    product_category = []
+
+    @products.find_all do
+      if product.category == category
+        product_category << product.category
+      end
+    end
+    product_category
+  end
+
+  def details
+    self.attributes
+
+  end
+
 end
--- a/lib/shopping_cart.rb
+++ b/lib/shopping_cart.rb
@@ -8,7 +8,15 @@ class ShoppingCart
   end
 
   def add_product(product)
-    @products << product 
+    @products << product
+  end
+
+  def total_number_of_products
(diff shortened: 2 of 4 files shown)


lib/product.rb
```diff
@@ -1,13 +1,19 @@
...
+    @quantity = quantity.to_i
+    @hoarded = false
   end
 
   def total_price
     @unit_price * @quantity
   end
+
+  def hoard
+    @hoarded = true
```
timomitchel: This is exactly what we were looking for here.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,38 @@
+class ShoppingCart
+  attr_reader :name, :capacity, :products
+
+  def initialize(name, capacity)
+    @name = name
+    @capacity = capacity.chomp('items').to_i
```
timomitchel: Nice job chomping this part of the string. Surprisingly just `capacity.to_i` will work here, because ruby loves you.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,38 @@
...
+    @products = []
+  end
+
+  def add_product(product)
+    @products << product
+  end
+
+  def total_number_of_products
+    @products.inject(0) {|result, product| result + product.quantity }
```
timomitchel: `inject` works here, but this reads better with the enumerable `sum`. It's designed specifically for this job.

lib/shopping_cart.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def is_full?
+    @products.inject(0) {|result, product| result + product.quantity } > @capacity
```
timomitchel: You duplicate work here. This method should read:

lib/shopping_cart.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def products_by_category(category)
+    product_category = []
+
+    @products.find_all do
```
timomitchel: I like that you chose `find_all` here, but you're using it like it's an `each` here. Remember that find_all returns a subset of the original array based on a condition. So it would read:
```
products.find_all {|product| product.category == category}
```

lib/shopping_cart.rb
```diff
@@ -0,0 +1,38 @@
...
+      if product.category == category
+        product_category << product.category
+      end
+    end
+    product_category
+  end
+
+  def details
+    self.attributes
```
timomitchel: Looks like you ran out of time here, but the hash we wanted to build could be accomplished with `self` or the instance variables:
```
{
name: name, 
capacity: self.capacity
}
```

test/product_test.rb
```diff
@@ -9,35 +9,38 @@ def test_it_exists
...
   def test_it_can_get_total_price
-    skip
     product1 = Product.new(:paper, 'toilet paper', 3.70, '10')
     product2 = Product.new(:meat, 'chicken', 4.50, '2')
 
     assert_equal 37.0, product1.total_price
     assert_equal 9.0, product2.total_price
   end
+
+  def test_product_is_hoarded
```
timomitchel: Nice job with the test, but remember that you lose points if you don't follow the interaction pattern naming exactly.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,62 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/product'
+require './lib/shopping_cart'
+
+class ShoppingCartTest < Minitest::Test
+
+  def setup
+    @cart = ShoppingCart.new("King Soopers", "30items")
```
timomitchel: Nice use of the setup method here.

test/shopping_cart_test.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    assert_equal 13, @cart.total_number_of_products
+  end
+
+  def test_cart_is_full
+    @cart.add_product(@product1)
+    @cart.add_product(@product2)
+    @cart.add_product(@product3)
+
+    assert_equal false, @cart.is_full?
```
timomitchel: Nice job testing both of the boolean values here.

